PR #37, opened 2017-09-11
Make logs searchable; Enable fpm slow log

tatyree:
It appears cloudwatch has very limited search capabilities for non-json
formatted logs–simple wildcards only. JSON formatted logs, however,
have a much more sophsiticated search DSL.  This should make it much
easier to find things in the logs.

I have also enabled the fpm slow log in an effort to track down the
cause of the remaining 504s and slow event feed load times.  The setting
I've chosen are aggressive–they may need review after a few days.  Also,
it is worth noting that triggering this log locally is tricky.  As a
result, I do not know its format, so it may be diffcult to search on
cloudwatch. There does not seem to be a simple way to configure it to
output json.  I will experiment with doing so if it becomes necessary.

### DIFF
--- a/docker/etc/php/7.1/fpm/pool.d/pool.conf
+++ b/docker/etc/php/7.1/fpm/pool.d/pool.conf
@@ -229,7 +229,7 @@ pm.max_requests = 1500
 ;       anything, but it may not be a good idea to use the .php extension or it
 ;       may conflict with a real PHP file.
 ; Default Value: not set
-;pm.status_path = /status
+; pm.status_path = /status
 
 ; The ping URI to call the monitoring page of FPM. If this value is not set, no
 ; URI will be recognized as a ping page. This could be used to test from outside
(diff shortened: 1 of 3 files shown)


docker/etc/php/7.1/fpm/pool.d/pool.conf
```diff
@@ -314,20 +314,20 @@ pm.max_requests = 1500
...
-;request_slowlog_timeout = 0
+request_slowlog_timeout = 1s
 
 ; The timeout for serving a single request after which the worker process will
 ; be killed. This option should be used when the 'max_execution_time' ini option
 ; does not stop script execution for some reason. A value of '0' means 'off'.
 ; Available units: s(econds)(default), m(inutes), h(ours), or d(ays)
 ; Default Value: 0
-;request_terminate_timeout = 0
+request_terminate_timeout = 5s
```
digitalronin: Part of me thinks this is overly aggressive, but OTOH when the endpoints work, they usually do so in <1s, so I guess it's OK. Just feels a bit scary!

PR #39, opened 2017-09-13
Gruntfile not finding source sass files

brown-a2:
@tatyree when you get a chance and you have a look at.

### DIFF
--- a/docker/Gruntfile.js
+++ b/docker/Gruntfile.js
@@ -1,7 +1,7 @@
 module.exports = function(grunt) {
   "use strict";
 
-  var srcDir = grunt.option('source-path') || 'web/app/themes';
+  var srcDir = grunt.option('source-path') || 'bedrock/web/app/themes';
   var themePath = '/mojintranet/';
 
   var jsSrcListBase = [


IruneItoiz:
Does this work if you build the image from scratch using this new gruntfile?

brown-a2:
Not sure, I was wondering that as well. one for Todd

docker/Gruntfile.js
```diff
@@ -1,7 +1,7 @@
 module.exports = function(grunt) {
   "use strict";
 
-  var srcDir = grunt.option('source-path') || 'web/app/themes';
+  var srcDir = grunt.option('source-path') || 'bedrock/web/app/themes';
```
tatyree: Which sass file is it failing to build? At which point of the build process?

tatyree:
@brown-a2 can you post a brief outline of the steps to take to reproduce the build failure including the error message and the particular file that it's failing to build.  If it's more than a couple of steps, can you open it as an issue on the project and link to this PR, please?

brown-a2:
Running grunt in the root of the `wp_local_dev` container (after it has been built) returns the following error below. 

```Running "sass:dev" (sass) task
>> Source file "web/app/themes/mojintranet/assets/css/src/style.scss" not found.
>> Source file "web/app/themes/mojintranet/assets/css/src/ie.scss" not found.
>> Source file "web/app/themes/mojintranet/assets/css/src/fonts.scss" not found.```

The error is solved by modifying the source path in the Gruntfile.js (line 4) to include the newly created /bedrock/ folder but this then breaks the container on build.

IruneItoiz:
This could be fixed if we move the grunt files inside the theme, as then it would be self contained?

PR #43, opened 2017-09-21
Comments added to pages and news posts

brown-a2:
This adds comments to all news posts. Also added to select pages; these include about, campaign content, tabs, media grid and default template pages.


wp-content/themes/mojintranet/page_media_grid.php
```diff
@@ -1,49 +1,60 @@
-<?php if (!defined('ABSPATH')) die();
-
+<?php if (!defined('ABSPATH')) {
```
IruneItoiz: You would usually have this on a single line, as it's just a way to prevent direct access.

PR #45, opened 2017-10-09
Ensure caching meets expectations of end users

tatyree:
The Content team expect that newly published articles will appear on the
site within 60 seconds of publication.  Due to the aggressive nature of
browser caching, and our varying `cache_timeout` settings, this was
frequently not the case.  This change ensures that all caching will be
done for 60 seconds.

I started by making all of the `cache_timeout` settings equal to 60,
then I realised this could be DRY'd out by simply removing them and
setting the default timeout to 60 seconds.  I replaced the
`cache_timeout` key for error pages, only-set to zero seconds, this
continues to prevent the error pages from caching by sending max-age=0;

This *may* cause performance problem on production, so we should monitor
server performance for some time after it is released.

### DIFF
--- a/wp-content/themes/mojintranet/inc/cache.php
+++ b/wp-content/themes/mojintranet/inc/cache.php
@@ -4,8 +4,8 @@
   }
 
   function dw_clear_headers() {
-    $cache_timeout = 90;
-    
+    $cache_timeout = 60;
+
     $headers['Cache-Control'] = 'public, max-age='.$cache_timeout;
     $headers['Expires'] = gmdate('D, d M Y H:i:s \G\M\T', time() + ($cache_timeout?:60));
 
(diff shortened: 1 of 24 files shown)


wp-content/themes/mojintranet/inc/cache.php
```diff
@@ -4,8 +4,8 @@
   }
 
   function dw_clear_headers() {
-    $cache_timeout = 90;
-    
+    $cache_timeout = 60;
```
brown-a2: not sure if it's helpful to make a small comment indicating that it's referring to seconds?

tatyree: Cache-Control header entries are always in seconds...

PR #110, opened 2018-03-07
Docker refactor

tatyree:
DO NOT MERGE.  

This is currently an advisory PR.


Dockerfile
```diff
@@ -0,0 +1,89 @@
...
+  && echo 'deb http://nginx.org/packages/mainline/debian/ stretch nginx deb-src http://nginx.org/packages/mainline/debian/ stretch nginx' > /etc/apt/sources.list.d/nginx.list \
+  && curl -fsSL http://nginx.org/keys/nginx_signing.key | apt-key add - \
+  && apt-get update \
+  && apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" -y install \
+  fuse \
+  git \
+  gnupg \
+  libffi-dev \
+  mariadb-client \
+  newrelic-php5 \
```
brown-a2: Does it matter here that we are running PHP7 not 5?

Gruntfile.js
```diff
@@ -0,0 +1,137 @@
+module.exports = function(grunt) {
+  "use strict";
+
+  var srcDir = grunt.option('source-path') || 'web/app/themes';
```
brown-a2: you know how the path to this file changes to 'bedrock/web/app/themes' once it's moved by the Dockerfile ... will any of this change in this refactor?

config/environments/staging.php
```diff
@@ -0,0 +1,7 @@
+<?php
```
brown-a2: I guess we can get rid of this environment now or change the dev to staging as you suggested.

moj.json
```diff
@@ -0,0 +1,33 @@
...
+
+    "ministryofjustice/dw-mvc":"dev-master",
+    "ministryofjustice/dw-document-revisions":"dev-update-composer-json",
+    "ministryofjustice/dw-live-drafts":"dev-update-composer-json",
+    "ministryofjustice/dw-markdown":"dev-update-composer-json",
+    "ministryofjustice/intranet-theme-clarity":"dev-master",
+    "ministryofjustice/intranet":"dev-master",
+    "ministryofjustice/wp-rewrite-media-to-s3": "*",
+    "ministryofjustice/like-button-for-wordpress":"dev-master",
+    "wpackagist-theme/twentyseventeen":"1.3"
```
brown-a2: "wpackagist-theme/twentyseventeen":"1.3" This line can go, not using this theme whatsoever.

Dockerfile
tatyree: It doesn't.  This is their standard package.  Technically, isn't 7 actually 5.7?

tatyree: Oh and this is the package we're using now, and we run 7.1

Gruntfile.js
tatyree: No.  That's for later.

config/environments/staging.php
tatyree: Good point.  I'll pull it.

moj.json
tatyree: I wondered...It's been in there forever, but didn't seem to be doing anything.

bedrock/config/application.php
```diff
@@ -0,0 +1,85 @@
...
+define('LOGGED_IN_SALT', env('LOGGED_IN_SALT'));
+define('NONCE_SALT', env('NONCE_SALT'));
+
+/**
+ * Custom Settings
+ */
+define('AUTOMATIC_UPDATER_DISABLED', true);
+define('DISABLE_WP_CRON', true);
+define('DISALLOW_FILE_EDIT', true);
+define('WP_DEFAULT_THEME', 'ppo');
```
ollietreend: `WP_DEFAULT_THEME` should be `intranet`, or whatever the directory name of your theme is. Although this only applies when the DB is blank, since the active theme is usually specified somewhere in the depths of the `wp_options` table.

brown-a2: I assume this would be our child theme name `intranet-theme-clarity` not parent theme?

ollietreend: Correct. It should be the directory name of whichever theme you want to be active.

ollietreend:
Just to be clear – you've actively chosen to include the dw-* custom plugins directly into this codebase, rather than having them split out into separate repos over GitHub and pulled in as dependencies. Is that correct?

tatyree:
Yes, the `dw-*` plugins have been deliberately moved into this project and their repos will be deprecated then removed after this goes to production.

tatyree:
No longer relevant. Closing.

PR #267, opened 2020-01-03
Colour checker

brown-a2:


web/app/themes/clarity/inc/admin/js/colour-contrast-checker.js
```diff
@@ -1,61 +1,53 @@
...
-                        }
-                    }
-                });
-            }
-            else {
-                hideContrastMessages(input);
+jQuery(function ($) {
+  $('.colour_check input').each(function () {
+    checkContrast($(this))
+  })
```
wilson1000-MoJ: Like most function end parenthesis, there is no semicolon at the end.

PR #269, opened 2020-02-05
Added Excerpt to campaign hub template

RobLoweMOJ:
Adds the excerpt custom field.


web/app/themes/clarity/page_campaign_landing.php
```diff
@@ -21,6 +21,19 @@
 
   <h1 class="o-title o-title--page"><?php the_title(); ?></h1>
 
+    <?php
+    $excerpt = get_field('dw_excerpt');
+
+    if (strlen($excerpt) > 0) {
```
brown-a2: Out of curiosity, could you not just have  `if ($excerpt)` rather then need to check the string length?

PR #270, opened 2020-02-10
User agency bug

brown-a2:
Agency Admins have the capability to add users and assign them to a
particular agency. Cat from LAA identified a bug where she was not able
to assign a new user to her agency. This fix addresses this bug.

Part of the problem was a legacy code issue, where Agency Admins could
select multipule agency views at once via checkboxes. Now this fix implements ratio
buttons allowing an agency admin editor the ability to only select one
agency view at a time.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,4 @@ There are several `make` commands configured in the `Makefile`. These are mostly
 
 ## Bedrock
 
-This project is based on Bedrock. Therefore, much of the Bedrock documentation will be applicable.
-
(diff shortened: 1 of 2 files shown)


RobLoweMOJ:
Looks Good. Though I would recommend double testing ( I know overkill haha) with the each of the different user roles as I seen the permissions can easily become buggy.

brown-a2:
Thanks, yes agreed! I've checked all roles and scenarios on my local machine but I'll do it all again once it's on dev. Maybe I'll get a few of you involved in that as well :)

PR #272, opened 2020-03-02
Typo fix

brown-a2:
Fixed "registration" typos in the email subject and body. While I was at it, I also updated the email addresses using the old intranet email address with the new team email.


RobLoweMOJ:
Looks good

PR #274, opened 2020-03-17
403 content update

brown-a2:
Update the 403 page to give staff more information on a common connection issue.

### DIFF
--- a/web/app/themes/clarity/error-pages/403.html
+++ b/web/app/themes/clarity/error-pages/403.html
@@ -447,7 +447,7 @@
     </ul>
     <h2>Troubleshooting</h2>
     <h3>You are on a MoJ network but still seeing the 403 forbidden page?</h3>
-    <p>Sometimes your browser cannot tell when you've switched to a secure MoJ network and will continue to display the 403 forbidden page. To fix this, once you are sure you are connected to an MoJ network/wifi or VPN, completely close down your browser and any open browser windows on your device. Then reopen the browser and try the intranet address again.</p>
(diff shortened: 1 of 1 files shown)


web/app/themes/clarity/error-pages/403.html
```diff
@@ -441,7 +441,8 @@ <h3>Please access from:</h3>
       <li>Within the network, at an office or commuter hub,</li>
       <li>Or from home, using VPN access.</li>
     </ul>
-    <p></p>
+    <h3>Troubleshooting</h3>
+    <p><strong>You are on a MoJ network but still seeing the 403 forbidden page?</strong> <br><br>Sometimes your browser cannot tell when you've switched to a secure MoJ network and will continue to display the 403 forbidden page. To fix this, once you are sure you are connected to an MoJ network/wifi or VPN, completely close down your browser and any open browser window on your device. Then reopen the browser and try the intranet address again.</p>
```
BNewing: Looks good! Could we change the `<br>` to CSS instead? Very minor thing, but they aren't the best for screen reader users: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br#Accessibility_concerns

BNewing: I'd also be tempted to change it from ' completely close down your browser and any open browser windows' to 'and any other open browser windows' - I skim read it at first and picked out the open bit and thought it meant re-open the browser twice. That might just be me though!

wilson1000-MoJ:
Hey @brown-a2 , a definite improvement content wise! Thank you. Hopefully it will slow down the emails in our inboxes 👍 all good from me.

web/app/themes/clarity/error-pages/403.html
brown-a2: yup, that makes sense

BNewing: Looks great :D

PR #287, opened 2020-04-22
User Deletion from frontend of site

RobLoweMOJ:


web/app/themes/clarity/src/components/c-delete-user/view.php
```diff
@@ -0,0 +1,102 @@
+<?php
+
```
brown-a2, line 2: so empty...

web/app/themes/clarity/src/components/c-comment-form/view.php
```diff
@@ -48,8 +48,10 @@
         comment_form($comments_args);
         ?>
       <p class="secondary-action">
-        <a href="https://intranet.justice.gov.uk/commenting-policy/">MoJ commenting policy</a>
+        <a href="https://intranet.justice.gov.uk/commenting-policy/">MoJ commenting policy</a><br/>
+          <a href="<?php echo get_bloginfo('url'); ?>/delete-account/">Delete your Account</a>
```
brown-a2, line 52: should this be hardcoded?

PR #299, opened 2020-08-28
Fix javascript compiling errors

wilson1000-MoJ:
Readjusted Laravel mix to remove babel and process javascript files efficiently
I found a script that loads all the JavaScript objects called `script-loader.js`. It's potentially a great little script as it loads objects only when they are needed. In line with ES6 I have converted the component scripts (as well as global scripts) to exportable modules and imported them in the `script-loader.js` file as dependencies for the loader. There are 2 benefits here:

- we don't need to maintain babel anymore
- we can keep the location of script dependencies in one place

I can't see any cons for this approach, the script-loader always needs updating when a component with a script is added which is why moving dependancies here works well.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -10,9 +10,6 @@ charset = utf-8
 trim_trailing_whitespace = true
 insert_final_newline = true
 
-[*.js]
-indent_size = 2
-
 [Makefile]
 indent_style = tab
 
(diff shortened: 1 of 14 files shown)


brown-a2 (approved):
Looks good to me. I think we should test some of these JS controlled functionality on the dev environment to make sure they are working, particularly the lefthand nav and tabs as they are critical to the site.

wilson1000-MoJ:
I'm testing as I go buddy :o)

PR #305, opened 2021-01-18
Intranet v3 cherrypick

brown-a2:
* Takes new banner code from our `elasticsearch` branch so that we can use it on intranet-v3 staging for testing and eventually in our final release.


RobLoweMOJ (approved):
👍

PR #316, opened 2021-12-15
Update composer and wordpress 5.8

MeenaModhvadia:
Updated all plugins
Updated wordpress version from 5.7


brown-a2 (approved):
Looks good :)

PR #320, opened 2022-01-26
Add Official Solicitor and Public Trustee (OSPT) agency to intranet

brown-a2:
Added in OSPT to all the various locations needed so that they can have their own independent area on the intranet.

TODO: once the agency has published everything and is ready to "go live" we need to add the link back into the switcher. This is done by removing the code I've added here and removing the legacy OSPT intranet link (in the same file).


web/app/themes/clarity/acf-json/group_5c10285516106.json
```diff
@@ -169,10 +169,8 @@
             },
             "admin_only": 0,
             "choices": {
-                "emergency": "Emergency (red border)",
-                "service": "Service update (orange border)",
-                "summary": "Summary (blue border)",
-                "information": "Information (black border)"
+                "emergency": "Emergency (red banner)",
+                "service": "Service update (orange banner)"
```
RobLoweMOJ, line 173: I would add these back in

RobLoweMOJ (approved):

PR #334, opened 2022-03-02
Remove duplicated headings

MeenaModhvadia:
- Headings changed from multiple H1's and H2 to H1, H2 and H3
- Span added to Time, Location and Hidden Date

### DIFF
--- a/web/app/themes/clarity/src/components/c-calendar-icon/view-eventsapi.php
+++ b/web/app/themes/clarity/src/components/c-calendar-icon/view-eventsapi.php
@@ -1,5 +1,5 @@
 <time class="c-calendar-icon" datetime="<?php echo $start_date . ' ' . $start_time; ?>">
-    <h2 class="u-visually-hidden">Date:</h2>
+    <span class="u-visually-hidden">Date:</span>
     <span class="c-calendar-icon--dow"><?php echo $get_day; ?></span>
     <span class="c-calendar-icon--dom"><?php echo $get_day_num; ?></span>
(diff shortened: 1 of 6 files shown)


web/app/themes/clarity/src/components/c-calendar-icon/view-eventsapi.php
```diff
@@ -1,5 +1,5 @@
 <time class="c-calendar-icon" datetime="<?php echo $start_date . ' ' . $start_time; ?>">
-    <h2 class="u-visually-hidden">Date:</h2>
+    <span class="u-visually-hidden">Date:</span>
```
brown-a2, line 2: Looks good but could you use the <time> element here?

brown-a2 (approved):

web/app/themes/clarity/src/components/c-event-listing/view.php
```diff
@@ -70,12 +71,14 @@
         ?>
 
       <div class="c-event-listing--time">
-        <h2>Time:</h2><?php echo $time; ?>
+        <span>Time:</span>
```
brown-a2, line 74: I'd maybe  use the `<time>` element here instead of `<span>`? So it would be `<time>Date:</time>` rather then `<span>Date:</span>`.

MalcolmVonMoJ: Using `<time>` is a good idea, but it would be the next line `<?php echo $time; ?>` that would have the tag - rather than replacing the `<span>`.

PR #336, opened 2022-03-16
Empty anchor link removed

MeenaModhvadia:
- Removed empty anchor link with a check if the url exists


brown-a2 (approved):
Just curious, does there need to be a check on the title as well?

brown-a2:
I can't imagine we would want to have just a title or just a link, you'd always want to have both. What do you think?

PR #342, opened 2022-03-30
Add Judicial Appointments Commission (JAC) agency

brown-a2:
* Changed the intranet data structure to allow a new agency. One thing to note, is that there was an old JAC entry here that deviated from the standard agency prefix, so instead of JAC it was the full judicial appointments commission title. So I've removed that who old entry and created a new one using just JAC. Now it is inline with how we approach the other agencies.
* Added JAC to the menu registry
* Added JAC header colour
* Added JAC email (and email to comment allowlist)
* Add ACF fields for agency homepage
* Removed JAC from appearing in switcher until site goes live


web/app/themes/clarity/inc/admin/acf-field-group.php
```diff
@@ -5904,37 +5879,10 @@
 			'ui_on_text' => '',
 			'ui_off_text' => '',
 		),
-		array(
```
RobLoweMOJ, line 5907: Why is agency_wide_banner_type being removed?

web/app/themes/clarity/inc/admin/acf-field-group.php
```diff
@@ -5951,26 +5899,6 @@
 			'append' => '',
 			'maxlength' => '',
 		),
-		array(
```
RobLoweMOJ, line 5954: why is agency_wide_banner_message being removed

web/app/themes/clarity/inc/mymoj.php
```diff
@@ -249,69 +249,7 @@ public static function get_apps($agency = 'hq')
             'external' => true,
           ],
         ],
-        'judicial-appointments-commission' => [
```
RobLoweMOJ, line 252: Should this be changed to JAC rather then removed?

RobLoweMOJ (commented):
Why in the ACF json is it replacing OSPT fields. Should these be new fields?

web/app/themes/clarity/acf-json/group_5c10285516106.json
```diff
@@ -492,7 +490,7 @@
                     {
                         "field": "field_5b696b695a61a",
                         "operator": "==",
-                        "value": "ospt"
+                        "value": "jac"
```
RobLoweMOJ, line 493: As mentioned seems to replacing ospt

web/app/themes/clarity/inc/admin/acf-field-group.php
brown-a2: Yes I noticed this and the OSPT fields being removed, it's like ACF reverted back to some previous state. I'm not sure how that is possible though, given it is a PHP file. I'm having a look now, thanks

web/app/themes/clarity/inc/admin/acf-field-group.php
brown-a2: 👍

web/app/themes/clarity/inc/mymoj.php
brown-a2: I opted to just create a new entry after the OSPT using the JAC prefix instead and then after go through and removed all references to this old one. I guess I could have replaced it just as well. I've found though with the intranet it is sometimes better to create a new entry and then when it is working go through and remove the old one.

web/app/themes/clarity/acf-json/group_5c10285516106.json
brown-a2: same as mentioned above

brown-a2:
I've manually added back in the ACF fields I could see have changed. Hopefully that's got them.

RobLoweMOJ (approved):
Some ACF changes on here would double check on staging.

PR #352, opened 2022-04-27
Non desc label2

MeenaModhvadia:
- Added fieldset and legend
- Removed colon from Date, Keyword and others that use filter
- Added hint text for News and Event filters
- Amended text to Filter By


MalcolmVonMoJ (approved):
Looks good

PR #355, opened 2022-05-04
News search

MeenaModhvadia:
- Added aria role for news search results to be announced
- Added focus for news search results
- Removed extra space from class definition


web/app/themes/clarity/page_campaign_landing.php
```diff
@@ -38,7 +38,7 @@
     ?>
   </div>
 
-  <div class="l-primary campaign">
+  <div class=" campaign">
```
brown-a2, line 41: small gap still there

brown-a2 (approved):

PR #358, opened 2022-06-17
Reflow

MeenaModhvadia:
- Function uses string replace to add new div around table with a class for tabbed content
- Function uses string replace to add new div added around table with with a class for rich text block content


MalcolmVonMoJ (approved):
Looks good to me

PR #365, opened 2022-08-31
Update various plugins plus fixed vulnerabilities

MeenaModhvadia:
- ElasticPress and various other plugins updated.
- Dependencies upgraded as per recommendations to remove vulnerabilities.


composer.lock
```diff
@@ -964,10 +965,10 @@
         },
         {
             "name": "koodimonni-language/core-en_gb",
-            "version": "5.9.3",
+            "version": "6.0.1",
```
wilson1000-MoJ, line 968: Hey Meena :) I notice the core language has been updated for WordPress; however, the core WP installation hasn't been updated. This constraint is managed in `composer.json` - it would be worth checking that WP can be updated to version 6.0.1 and that the WP language core and core versions match.

wilson1000-MoJ (changes requested):

wilson1000-MoJ (approved):
Looks great, well done!

PR #367, opened 2022-09-27
Fix checksum verification error

wilson1000-MoJ:
Since the last build took place, the plugin ACF version 5.12.3 was removed from the source and replaced. This action caused the composer shasum to be invalid.

This PR updates the ACF shasum. 

```
The checksum verification of the file failed (downloaded from https://composer.wp.dsd.io/dist/acf/advanced-custom-fields-pro/acf-advanced-custom-fields-pro-5.12.3.zip)
```


Polsonby (approved):
LGTM :)

PR #369, opened 2022-10-05
CDPT-269 : create an Action to check composer can run successfully

wilson1000-MoJ:
An action to run composer checks. Makes sure we have no errors.

### DIFF
--- a/.github/workflows/php.yml
+++ b/.github/workflows/php.yml
@@ -1,0 +1,39 @@
+name: PHP Composer
+
+on:
+  push:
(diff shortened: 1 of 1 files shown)


wilson1000-MoJ:
Line 34 exits with a failure on `run: composer validate --strict`. This occurs (potentially) due to the warnings around no versions being present. In the new version for php8.1, all version numbers are specified. Removing `--strict` to allow for this.

rohanssalunkhe (approved):

PR #375, opened 2022-11-03
Update view.php

rohanssalunkhe:


Polsonby (approved):
LGTM :)

wilson1000-MoJ:
URL doesn't load, but (as Rohan pointed out) this may well be due to the use of a VPN.

PR #379, opened 2022-11-11
Update readme with NodeJS requirement

digitalali-moj:
This is me testing if I can push stuff, but also just adding that you need to have a supported version of Node installed. Otherwise you get the error that looks like this when trying to install.

library: 'digital envelope routines',
  reason: 'unsupported',
  code: 'ERR_OSSL_EVP_UNSUPPORTED'


Polsonby (approved):
👍

PR #380, opened 2022-12-08
CDPT-217 Remove dev Tagman that doesn't work anyway

digitalali-moj:
Tagman production is working but Dev wasn't from what I could see on GA. Removing for simplicity.


wilson1000-MoJ (approved):
Cool :)

PR #396, opened 2023-04-25
CDPT-678 : Integrate GovUK Notify

wilson1000-MoJ:
We rely on one environment variable stored in our params library for this update.

SMTP_PASSWORD will store the GovUK Notify API key. Once this is updated, the service will be active.


Polsonby (approved):
LGTM, nice one 👍

PR #414, opened 2023-05-17
CDPT-689 production release for php8.2 upgrade

wilson1000-MoJ:


Dockerfile
```diff
@@ -1,25 +1,13 @@
...
-    echo 'deb http://apt.newrelic.com/debian/ newrelic non-free' > /etc/apt/sources.list.d/newrelic.list && \
-    curl -fsSL https://download.newrelic.com/548C16BF.gpg | apt-key add - && \
-    mv docker/conf/nginx/server.conf /etc/nginx/sites-available/ && \
-    mv docker/conf/php-fpm/newrelic.ini /etc/php/7.4/fpm/conf.d/ && \
-    mv docker/init/configure-maintenance-mode.sh /etc/my_init.d/ && \
-    chmod +x /etc/my_init.d/configure-maintenance-mode.sh && \
-    apt-get update && \
-    apt-get install -y libffi-dev newrelic-php5 && \
-    rm -f /etc/my_init.d/configure-ip-whitelist.sh
+RUN mv docker/conf/nginx/server.conf /etc/nginx/sites-available/
```
digitalali-moj, line 10: Does this mean we're not using PHP FPM now?

web/app/themes/clarity/src/components/c-input-container/on-change.js
```diff
@@ -0,0 +1,7 @@
+export default (function ($) {
```
digitalali-moj, line 1: Just checking if this is part of the search enhancement and if it should go in with this.

digitalali-moj (approved):
Excellent work, and a lot! Some ignorant comments left. Would be nice to talk through bits like FPM, buildspec.yml to understand a bit better!

Dockerfile
wilson1000-MoJ: Hey Jav. All deleted lines are included in our base image; intranet-base:latest

Makefile
```diff
@@ -18,20 +18,24 @@ docker-clean:
 	bin/docker-clean.sh
 
 # Run the application
-run:
-	docker-compose up
+run: dory
```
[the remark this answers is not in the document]
wilson1000-MoJ: Not in this case mate; dory references a make command on line 33.

wilson1000-MoJ:
Thanks so much. Let's have a catch up tomorrow.

web/app/themes/clarity/src/components/c-input-container/on-change.js
wilson1000-MoJ: Good question. 
This is to do with supporting radio elements on the search form. I've added it to compliment the form builder api which we will be using in the next couple of weeks so left it in for that. Happy to walk through what I did tomorrow, if you're free buddy.

PR #419, opened 2023-05-30
Fix undefined array key

wilson1000-MoJ:


vertism:
Is there a test to show what this fixes, or some more context on what the original issue was?

wilson1000-MoJ:
> Is there a test to show what this fixes, or some more context on what the original issue was?

Yes, apologies. Ticket CDPT-734:
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?modal=detail&selectedIssue=CDPT-734

The code that was updated:
https://github.com/ministryofjustice/like-button-for-wordpress/commit/a67d6e0b2c24e82c4283429abbc08fe3bf867c41

In this PR, composer pulls the new code from the above plugin.

wilson1000-MoJ:
**A little more context:**
The short ternary used in the `like-button-for-wordpress` plugin was not evaluating `isset`. This resulted in the warning that an array key was undefined.  The operator `??` is used instead. This operator will check for `isset` and return the value if it is set, or empty array if not.

vertism (approved):
Thanks for the context! LGTM

PR #420, opened 2023-05-30
CDPT-732 Fix undefined variable $context

wilson1000-MoJ:
**Ticket:**
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?modal=detail&selectedIssue=CDPT-732

**Context:**
Regions will (on occasion) fire an error stating that $context is undefined. I couldn't replicate the error; however, I did find a line of code that was removed in a previous PR that causes this issue:

https://github.com/ministryofjustice/intranet/commit/b4b2ee10e1e649ea9fa3c3a09f5f3b7b03254b6f#diff-5cf875bf46fc9435a41322a66ebc892638af92996a023b7d676e5ba1a318845dL254

The trouble is, without spending time tracing this code, I cannot determine what the removal of this line did or why it was done. The only impact I can determine at this stage is that the removal of the line of code is causing this issue, as the PR detailed above doesn't offer any information about why it was removed.

We should move this PR to staging and test the impact there.

### DIFF
--- a/web/app/themes/clarity/inc/admin/agency_taxonomies/taxonomies/agency-taxonomy.php
+++ b/web/app/themes/clarity/inc/admin/agency_taxonomies/taxonomies/agency-taxonomy.php
@@ -11,9 +11,9 @@ abstract class Agency_Taxonomy extends Taxonomy
      * The agency the taxonomy belongs to
      * This should match the agency slug
      *
-     * @var string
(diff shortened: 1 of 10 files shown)


wilson1000-MoJ:
I am ditching to make it easier to understand the context.

PR #422, opened 2023-05-31
CDPT-731 Test array offset with fallback URL on false

wilson1000-MoJ:
**Ticket:**
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?modal=detail&selectedIssue=CDPT-731\

**Context:**
Some users will not allocate themselves an avatar image. This causes the co-author plugin to throw a warning, rendering the user with a broken image.

This update checks if a URL exists. If not, the code will request a default image based on their email address from Gravatar or the default Gravatar 'mystery person avatar.

http://en.gravatar.com/site/implement/images/


digitalali-moj (approved):
LGTM!

PR #424, opened 2023-05-31
CDPT-752 Remove alien character

wilson1000-MoJ:
**Ticket:**
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?modal=detail&selectedIssue=CDPT-752

**Context:**
The code block where this error has been identified is called (or should be) when a form request is performed more than once. It should prevent multiple form submissions.

At some point in the past an alien character was introduced however, the git history was erased by a team member which makes it impossible to identify the root cause.

### DIFF
--- a/web/app/themes/clarity/inc/content-filter/search.php
+++ b/web/app/themes/clarity/inc/content-filter/search.php
@@ -1,4 +1,5 @@
 <?php
+
 use MOJ\Intranet\Agency;
 use MOJ\Intranet\EventsHelper;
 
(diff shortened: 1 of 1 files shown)


web/app/themes/clarity/inc/content-filter/search.php
```diff
@@ -132,28 +133,28 @@ function load_search_results()
 
 function load_next_results()
 {
-    if (! wp_verify_nonce($_POST['nonce_hash'], 'search_filter_nonce')) {
-        ¦exit('Access not allowed.');
```
wilson1000-MoJ, line 136: This is the line that was updated to fix the issue

digitalali-moj (approved):
Thanks for cleaning up!

PR #435, opened 2023-06-28
Adjust left-hand menu for regions

wilson1000-MoJ:
Added regions to the custom menu ordering script.

**_Note entries below Documents:_**

> Regions are only visible when HMCTS view the admin panel. _Notes from Antonia_ is only visible when MoJ HQ are viewing; hence _Notes from Antonia_ is not visible in this menu example.

Screenshot 2023-06-28 at 11 58 19


MeenaModhvadia (approved):
Great work!

PR #441, opened 2023-07-24
Fix the campaign template pages

wilson1000-MoJ:
Defaults are not represented correctly. Explicitly set them.


digitalali-moj (approved):
Cheers!

PR #448, opened 2023-08-30
CDPT-903 Fix TypeError array_shift()

wilson1000-MoJ:
**_Context_**
This error occurs when a user is created or edited without a default agency being chosen. The code always expects an array to be passed; this is highly opinionated and isn't always true. The default 'hq' agency is given as a string when an agency has not been selected. 

The error this behaviour gives is fatal. It is destructive because the user's journey is halted immediately, and no reason is given to explain what happened.

**_The Fix_**
We assess an array and treat it accordingly. The code expects a string result on line 206 regardless of the origin type. In addition, it is assumed that the new or edited user should be created under the current context. Currently, a user is given HQ status if an agency isn't chosen. This release defaults the user to the current agency if one has not been selected.


digitalali-moj (approved):
Thanks for sorting this out.

PR #451, opened 2023-09-07
Configure radio buttons for HMCTS

wilson1000-MoJ:
**Context**
https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?selectedIssue=CDPT-918

Testing search impact using radio buttons for HMCTS has now completed. This PR activates radio buttons for HMCTS only. Once we have feedback from Intranet users, we can further consider release to the broader intranet audience.

### DIFF
--- a/web/app/themes/clarity/src/components/c-search-results-filter/view.php
+++ b/web/app/themes/clarity/src/components/c-search-results-filter/view.php
@@ -18,30 +18,68 @@ $activeAgency = $oAgency->getCurrentAgency();
 ?>
 <!-- c-search-results-filter starts here -->
 <section class="c-search-results-filter c-content-filter">
-    <p>You are searching across <strong><?php esc_attr_e($activeAgency['abbreviation']); ?></strong>. To search another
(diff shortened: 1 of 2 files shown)


web/app/themes/clarity/src/components/c-search-results-filter/view.php
```diff
@@ -18,30 +18,68 @@
...
+                'js-radios-onChange',
+                false,
+                null,
+                [
+                    ['All', 'any', $any],
+                    ['Pages', 'page', $page],
+                    ['Documents &amp; forms', 'document', $document],
+                    ['News', 'news', $news],
+                    ['Blogs', 'post', $post],
+                    ['Events', 'event', $event]
```
vertism, line 53: I noticed that the options are in alphabetical order below but not here. Is that intentional?

wilson1000-MoJ: Good Q. Yes, it is we ran this in the test, and the feedback was better.
Thanks for asking

vertism (approved):
LGTM

PR #465, opened 2024-01-26
Fix campaign categories being overwritten by different agencies

EarthlingDavey:
Fix campaign categories being overwritten by users from different agencies.

Add a function called `has_attached_terms_not_in_context` to catch if a post has a category attached that does not relate to current agency context.

Add a property called `run_has_attached_terms_not_in_context` to enable this on a taxonomy basis. Currently enabled for `Campaign_Category`, could also be enabled for other taxonomies that extend `Content_Category` e.g. `News_Category`.

Resolves: https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?selectedIssue=CDPT-1253

### DIFF
--- a/web/app/themes/clarity/inc/admin/agency_taxonomies/taxonomies/campaign-category.php
+++ b/web/app/themes/clarity/inc/admin/agency_taxonomies/taxonomies/campaign-category.php
@@ -46,4 +46,7 @@ class Campaign_Category extends Content_Category
             'assign_terms' => 'assign_campaign_categories',
         ),
     );
+
+    protected $run_has_attached_terms_not_in_context = true;
+
 }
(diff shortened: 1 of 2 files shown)


EarthlingDavey:
**To reproduce the bug:**

As HQ, go to a blog post that has another agency, the campaign categories available to the other agency must be less than for HQ e.g. locally http://intranet.docker/wp/wp-admin/post.php?post=473157&action=edit

Check a Campaign Category that is available to HQ but not OPG, e.g. 'Be open to Innovation updates'. Save the page.

Switch agency to OPG and [edit the same page](http://intranet.docker/wp/wp-admin/post.php?post=473157&action=edit). Check any Campaign Category , verify that the campaign category from the previous step is not shown in the list 'Be open to Innovation updates'. Save the page.

Switch back to HQ, you will see that 'Be open to Innovation updates' has been unchecked.

**To reproduce the fix:**

Run steps 1 & 2, at step 3 you should see that the side box for campaign category is not shown on the right hand side.

wilson1000-MoJ (approved):

PR #466, opened 2024-01-26
Debug helper function. Add log to file.

EarthlingDavey:
Add Debug::log to allow for debugging when writing to a log file is more suitable than printing to the page.


EarthlingDavey:
Please can we keep this issue as a placeholder.

@wilson1000-MoJ 's advised about not using local file system, and his suggested to check the stout.

PHP error messages are sent to stout (along with a log of network requests), therefore I think it is probably better for `WP_DEBUG_LOG` to never be set. 

I will document how to follow and possibly filter network request from stout, this will be useful for debugging PHP functions locally.

PR #518, opened 2024-04-11
Remove curl

wilson1000:
We have shaved 5 MB off the image after removing CURL!
I have also placed elaborate comments in the Dockerfile to help identify when new images are being created.

The Dockerfile has been cleaned, `chown nginx` user has been removed where the user was already nginx.

Screenshot 2024-04-11 at 16 16 10


EarthlingDavey:
Looking good ! :D

PR #522, opened 2024-04-12
Feature/cdpt 1487 intranet migration replace internal requests

EmilyHazlehurst:


public/app/themes/clarity/inc/content-filter/search.php
```diff
@@ -75,55 +75,81 @@ function load_events_filter_results()
...
+    $after = '';
+
+    if ($valueSelected) {
+        preg_match('/&after=([^&]*)&before=([^&]*)/', $valueSelected, $matches);
+        $after = date('Y-m-d', strtotime($matches[1]));
+        $before = date('Y-m-d', strtotime($matches[2]));
+    }
+
+    $nextPageToRetrieve = isset($_POST['nextPageToRetrieve']) ?  sanitize_text_field($_POST['nextPageToRetrieve']) : '';
+    $offset = $nextPageToRetrieve ? ($nextPageToRetrieve - 1) * $post_per_page : 0;
```
wilson1000, line 94: Hi @EmilyHazlehurst, first, well done on the work here, I just wanted to check; has this been tested locally?

There has been quite some update so I wanted to check for errors before reviewing and this line flagged up in Sentry.

[TypeError Unsupported operand types: string - int](https://ministryofjustice.sentry.io/issues/5198690924/?environment=development-damien&project=4506870789505024&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=0)

I would suggest wrapping the calculation in parenthesis to fix it but as always, will leave the decision up to you 👍 

```php
$offset = $nextPageToRetrieve ? (($nextPageToRetrieve - 1) * $post_per_page) : 0;
```

public/app/themes/clarity/inc/pagination.php
```diff
@@ -1,30 +1,35 @@
...
+      'post_type' => 'posts' ? 'post' : $type,
+      'post_status' => 'publish',
+      'tax_query' => [
+        'relation' => 'AND',
+        [
+          'taxonomy' => 'agency',
+          'field' => 'term_id',
+          'terms' => $activeAgency['wp_tag_id']
+        ],
+        ...($category_id ? [
```
wilson1000, line 21: It is sometimes uncertain what the ... (splat) operator is doing, which could make the code complex to read and debug in future.

Maybe; could a comment be left to explain the working or indeed, move the ternary out of the default array and push the values onto the array, if the category_id is set?

You're call 👍

public/app/themes/clarity/inc/pagination-newscategory.php
```diff
@@ -6,41 +6,47 @@ function get_newscategory_pagination($category_id)
...
+          ],
+          ...( $category_id ? [
+            'taxonomy' => 'news_category',
+            'field' => 'category_id',
+            'terms' =>  $category_id,
+          ] : []),
+      ]
+    ];
+
+    $query = new WP_QUERY($args);
```
wilson1000, line 30: Can we check this?
The call follows the constant convention, as in all uppercase, yet, it is acting like a class.

The class I think you need is `WP_Query()` - which in turn might have some drawbacks.

The class resets the post object when called. which means generally, we need to reset data after getting our results... this may not matter here but wanted to flag that just in case.

```php
<?php
   $temp_post = $post; // Storing the object temporarily
   $my_query = new WP_Query();
   while($my_query->have_posts()) {
      // Loop in here
   }
   $post = $temp_post; // Restore the value of $post to the original
?>
```

public/app/themes/clarity/src/components/c-article-item/view-blog-feed.php
```diff
@@ -49,15 +53,13 @@
     <div class="content">
         
         <h1>
-            <a href="<?php echo esc_url(get_permalink($id)); ?>"><?php echo $post['title']['rendered']; ?></a>
+            <a href="<?php echo esc_url($link); ?>"><?php echo $post->post_title; ?></a>
```
wilson1000, line 56: Maybe we could use: `get_the_title($id)`

This function will apply sanitation including language support, if needed.

public/app/themes/clarity/src/components/c-article-item/view-blog-feed.php
```diff
@@ -67,7 +69,7 @@
             ?>
 
         <div class="c-article-excerpt">
-            <p><?php echo $post['excerpt']['rendered']; ?></p>
+            <p><?php echo $post->post_excerpt; ?></p>
```
wilson1000, line 72: `$post['excerpt']['rendered'];`

I think 'rendered' is a modified value that has been sanitised - we need to trace the include path back to understand if this is the case though 😞 sorry about that.

As mentioned, we could use `get_the_excerpt()`

wilson1000:
Hi @EmilyHazlehurst, I hope you're doing ok 🙂 
There have been changes to the application that prevents access internally. This means all current API requests made internally are failing.  Will this update get through today?

EmilyHazlehurst:
> Hi @EmilyHazlehurst, I hope you're doing ok 🙂 There have been changes to the application that prevents access internally. This means all current API requests made internally are failing. Will this update get through today?

Hi @wilson1000, it should do - I'm working on it at the moment. I'll let you know if there are any issues.

public/app/themes/clarity/inc/content-filter/search.php
EmilyHazlehurst: Hi @wilson1000, I've updated this to use your suggestion. 

I have tested this locally; oddly I don't seem to get the error you described. 
Where do you generally look for error messages? I would have expected them to be printed to the page or the php container logs, but perhaps I'm missing a testing step that's Wordpress specific.

public/app/themes/clarity/inc/pagination-newscategory.php
EmilyHazlehurst: Thanks @wilson1000, I've updated this and added `wp_reset_postdata();` where I've used it as recommended in the docs, but let me know if I've missed something.

public/app/themes/clarity/inc/content-filter/search.php
wilson1000: Hiya.
It's in the link, it fired in Sentry:
https://ministryofjustice.sentry.io/issues/5198690924/?environment=development-damien&project=4506870789505024&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=14d&stream_index=4

wilson1000: ... that means the software running in the container is firing the error. When developing,  try to always run the containers as this will give you a preproduction environment to dev in - if you have your sentry ID set up in `.env` you will get an application error in Sentry 👍

PR #525, opened 2024-04-16
cloudfront-input secret for terraform. A/B public keys allows for key rotation with grace period.

EarthlingDavey:
This PR creates a `cloudfront-input` secret.

That means that we don't need to have our trusted public keys hardcoded in out `cloudfront.tf` file.

A change to `deploy.yaml` will allow us to rotate the trusted public keys with a grace period. Meaning that during key rotation we do not need to update the PEM values in our app and CloudFront at exactly the same instant. And, visitors with recently signed cookies will still have access to assets for a short while.

There are also a couple of typo fixes.


wilson1000 (approved):
Looks good

PR #526, opened 2024-04-16
Add wp offload media license

wilson1000:
Passes the new WP Offload Media Pro license key through to deployments


wilson1000:
> this guide

wilson1000:
> To get the package will you follow something like [this guide](https://deliciousbrains.com/wp-migrate-db-pro/doc/installing-via-composer/) or use an MoJ composer repository?

Hey @EarthlingDavey - we have the API key stored in this repo's secret. We can install the pro version and then use the key to activate it, just like we do with Notify. There's no need to use composer to authenticate.

EarthlingDavey:
I don't mind, I can make an entry in [pp-satis-config](https://github.com/ministryofjustice/pp-satis-config). 

I suppose the only downside of doing it that way is... we need to edit the repo for every version update.

wilson1000:
> we need to edit the repo for every version update.

Yes, it's a terrible downside. The trouble we have right now is the procurers haven't sent us login creds for WP Offload Media. If we use the method you suggest we can mitigate updating the repo... I like that a lot better 👍

EarthlingDavey:
Hi @wilson1000 - sorry about the delay. Trying to trouble shoot the broken `.tf` builds.

What do you reckon, should this be on hold until we have logins or a composer username/password pair from the procurers?

PR #530, opened 2024-04-17
CDPT-887 Remove unused error throw block.

EarthlingDavey:


deploy/development/deployment.tpl
```diff
@@ -67,7 +67,7 @@ spec:
               valueFrom:
                 secretKeyRef:
                   name: cloudfront-output
-                  key: cloudfront_public_key_ids
+                  key: cloudfront_public_keys
```
wilson1000, line 70: Is this reverted?

public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
```diff
@@ -128,11 +128,6 @@ public function getCloudfrontPublicKeyId(): string
         // Get the first 8 characters of the sha256.
         $public_key_short = substr($public_key_sha256, 0, 8);
 
-        // If the public key is not found, throw an exception.
-        if (empty($public_key_ids_and_keys)) {
-            throw new \Exception('AWS_CLOUDFRONT_PUBLIC_KEYS_OBJECT was not found');
-        }
-
```
wilson1000, line 135: Will we suffer with this removal or, was it a false positive?

EarthlingDavey: It's a false positive, it was left in after renaming some variables - it would always throw because $public_key_ids_and_keys is never set.

PR #538, opened 2024-04-18
CDPT-1481: Add JAC to the intranet switcher

EmilyHazlehurst:
- Remove JAC from intranet switcher ignore list


wilson1000 (approved):
Looks good Emily 👍

PR #550, opened 2024-04-26
CDPT-1528: Offload media

EmilyHazlehurst:
- Try to use basic auth in auth header directly

### DIFF
--- a/public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
+++ b/public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
@@ -51,7 +51,7 @@ class AmazonS3AndCloudFrontSigning
             if ($args['headers'] ?? false) {
                 $user_agent = $args['headers']['user-agent'] ?? false;
                 if ($user_agent === 'wp-offload-media') {
-                    $args['headers']['Authorization'] = 'Basic ' . $_ENV['BASIC_AUTH'];
+                    $args['headers']['Authorization'] = 'Basic ' . base64_encode($_ENV['BASIC_AUTH']);
                 }
             }
             return $args;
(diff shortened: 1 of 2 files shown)


public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
```diff
@@ -48,14 +48,12 @@ public function __construct()
...
 
         add_filter('http_request_args', function($args){
-            $headers = $args['headers'] ?? false;
-            if ($headers) {
-                $user_agent = $headers['user-agent'] ?? false;
+            if ($args['headers'] ?? false) {
+                $user_agent = $args['headers']['user-agent'] ?? false;
                 if ($user_agent === 'wp-offload-media') {
-                    $args['cookies'] = $this->createSignedCookie($this->cloudfront_url . '/*');
+                    $args['headers']['Authorization'] = 'Basic ' . $_ENV['BASIC_AUTH'];
```
EarthlingDavey, line 54: Hey, even if the secret looks like base64 when you run 

`kubectl -n $NSP get secret intranet-basic-auth -o yaml `

```
apiVersion: v1
data:
  auth: aW50cm...
```

k8s will decode it to user:xyz...

I think here $_ENV['BASIC_AUTH'] you might still need to run base64_encode.

EarthlingDavey (changes requested):

public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
EarthlingDavey, line 1: @EmilyHazlehurst 
To make sure this is working, could we also get $url as the 2nd function argument.

After the line `$args['headers']['Authorization']`, can we `error_log('url: ' . $url)` and `error_log(print_r($args, true))`.

Then we'll be able to check for 100% that the loopback is working. Hopefully the url will be `https://dev.intranet.justice.gov.uk/` meaning we won't have to worry about port 80 on Cloud Platform.

PR #553, opened 2024-04-29
Testing seperate auth and signed cookie

EmilyHazlehurst:


public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
```diff
@@ -51,10 +51,14 @@ public function __construct()
             if ($args['headers'] ?? false) {
                 $user_agent = $args['headers']['user-agent'] ?? false;
                 if ($user_agent === 'wp-offload-media') {
-                    $args['headers']['Authorization'] = 'Basic ' . base64_encode($_ENV['BASIC_AUTH']);
-                    error_log('url: ' . $url);
-                    error_log(print_r($args, true));
                     $args['cookies'] = $this->createSignedCookie($this->cloudfront_url . '/*');
+                    error_log('ua=offloadmedia: ' . $url);
+                    error_log(var_dump($args));
```
EarthlingDavey, line 56: I'm not sure, does var_dump work inside error_log? 

I have a feeling it might echo the value out as a response, instead of making it available a s a string for error_log.

EmilyHazlehurst: You're right, I've fixed it - thanks :)

PR #554, opened 2024-04-29
Get the unhashed password from github secrets

EmilyHazlehurst:


deploy/development/deployment.tpl
```diff
@@ -91,7 +91,7 @@ spec:
             - name: BASIC_AUTH
               valueFrom:
                 secretKeyRef:
-                  name: intranet-basic-auth
```
EarthlingDavey, line 94: If there's any reference to this in the codebase it will need to be changed. I think I've seen reference to it in ingress on an nginx config. Maybe also worth checking cloud-platform-environments / intranet-dev.

.github/workflows/deploy.yml
```diff
@@ -45,10 +45,13 @@ jobs:
           AWS_CLOUDFRONT_PUBLIC_KEY: "${{ secrets.AWS_CLOUDFRONT_PUBLIC_KEY_A }}"
           AWS_CLOUDFRONT_PRIVATE_KEY: "${{ secrets.AWS_CLOUDFRONT_PRIVATE_KEY_A }}"
           # AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING: "${{ secrets.AWS_CLOUDFRONT_PUBLIC_KEY_B }}"
+          BASIC_AUTH_USER: ${{ secrets.BASIC_AUTH_USER }}
```
EarthlingDavey, line 48: Don't forget to set these values via GitHub UI :)

deploy/development/secret.tpl
```diff
@@ -34,3 +34,11 @@ type: Opaque
...
   AWS_CLOUDFRONT_PUBLIC_KEY: "${AWS_CLOUDFRONT_PUBLIC_KEY_BASE64}"
   # AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING: "${AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING_BASE64}"
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: basic-auth-secret
+type: Opaque
+data:
+  auth: "${BASIC_AUTH_BASE64}"
```
EarthlingDavey, line 44: This will be the base64 encoded string with the hashed password.

On justice dev I didn't need the un-hashed password in the application. I think the user and pass should be added as 2 new lines around line number 19 in this file.

Then they'll be available as env vars to php :)

.github/workflows/deploy.yml
EmilyHazlehurst: Thanks for the reminder :)

deploy/development/deployment.tpl
EmilyHazlehurst: I've reverted this since I'm just using the username and password in the php

PR #559, opened 2024-06-25
Little updates

wilson1000:
1. Dockerfile line consolidation
2. Doubled PHP limits for upload and memory
3. Changed `wp-cli` path to relative
4. Linted 1 PHP file


EarthlingDavey (approved):
Looks good 👍

PR #560, opened 2024-06-27
Add options page and ACF fields for Prior Party Banner.

EarthlingDavey:
@wilson1000 , this includes a slight rename, to make the options page more flexible. Naming it to Prior Political Parties so as not to limit it to just Banner options. Do you reckon that sounds ok?


wilson1000 (approved):
Nice :)

PR #571, opened 2024-07-08
CDPT-1773 - Prior party banner add comments, end date inclusive.

EarthlingDavey:
This PR addressed the view of the Prior Party Banner:

- Comment the code
- Make end date inclusive, to match the admin WP_Query
- Improve repeater fields instructions
- Only use time_context when a banner's end date has not been passed
- Update logic around checking for editor or filtering out inactive banners
    As time_context is not always set for editors, use show_inactive query string.
- Fix shortcodes, render in correct place and with correct spacing.

### DIFF
--- a/web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
+++ b/web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
@@ -262,7 +262,7 @@ class PriorPartyBannerAdmin
 
                     // links
                     $link_admin = get_edit_post_link($post->ID);
-                    $link_view = get_permalink($post->ID) . '?show-inactive' . $link_view_suffix;
+                    $link_view = get_permalink($post->ID) . '?show_inactive' . $link_view_suffix;
 
                     // latest event
                     $event_data = $this->getTrackedDisplayString($post->ID);
--- a/web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
+++ b/web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
@@ -213,7 +213,7 @@ class PriorPartyBannerAdmin
 
             // If the end date has not passed... for previewing we need to set a time context after the end date.
             // i.e. exactly 00:00:00 the next day.
-            $link_view_suffix = $stop > new \DateTime() ? '?time_context=' . $stop->modify('+1 day')->format('U') : '';
+            $link_view_suffix = $stop > new \DateTime() ? '&time_context=' . $stop->modify('+1 day')->format('U') : '';
 
             // display the banner
             echo '<div class="prior-party-banner">
@@ -262,7 +262,7 @@ class PriorPartyBannerAdmin
 
                     // links
                     $link_admin = get_edit_post_link($post->ID);
-                    $link_view = get_permalink($post->ID) . $link_view_suffix;
+                    $link_view = get_permalink($post->ID) . '?show-inactive' . $link_view_suffix;
 
                     // latest event
                     $event_data = $this->getTrackedDisplayString($post->ID);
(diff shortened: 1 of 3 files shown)


web/app/themes/clarity/inc/admin/prior-party/prior-party-banner.php
```diff
@@ -80,6 +80,20 @@ public function init(): void
         $this->loadBanners();
     }
 
+    /**
+     * Get the preview time from the query string.
+     * 
+     * Validates the time_context query var against known end epochs.
+     * Returns false if if:
```
wilson1000, line 87: Double if -> `if if`

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
```diff
@@ -212,6 +212,19 @@ public function page(): void
             $stop = new \DateTime($this->banner["end_date"]);
 
 
+            // Init an array to hold the query string for viewing the banner.
+            $link_view_queries = [];
+            // If the banner is inactive, we need to set a query string to show it.
```
wilson1000, line 217: I'm not sure this is true @EarthlingDavey - we can see inactive banners on the front end using the `time_context` - what is `show_inactive` used for? Are we changing logic?

wilson1000: Also, `show_inactive` doesn't mean anything to an end user that would see the URL. 
If you are using this in replacement of `time_context` would a name like `preview_prior` be more understandable?

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
```diff
@@ -259,7 +272,7 @@ public function page(): void
 
                     // links
                     $link_admin = get_edit_post_link($post->ID);
-                    $link_view = get_permalink($post->ID) . '?time_context=' . $stop->format('U');
+                    $link_view = get_permalink($post->ID) . '?' . http_build_query($link_view_queries);
```
wilson1000, line 275: What would the expected outcome of this be?

My understanding either is:
```php
?show_inactive=''
```
or,
```php
?time_context=123456789
```

Is this right?

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
```diff
@@ -212,6 +212,19 @@ public function page(): void
             $stop = new \DateTime($this->banner["end_date"]);
 
 
+            // Init an array to hold the query string for viewing the banner.
+            $link_view_queries = [];
+            // If the banner is inactive, we need to set a query string to show it.
+            if(!$this->banner['banner_active']) {
+                $link_view_queries['preview_unpublished'] = '';
```
wilson1000, line 219: Brilliant result for the name 😄

wilson1000 (approved):

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
EarthlingDavey: The thing is, I think it was a bug that we set `time_context ` on every banner link.

It was meant to be a 'time travel forward' past the banner's end date. If the current time is past the banner's end date then time_context isn't really needed.

The logic is pretty much the same, but instead of inferring from `time_context` that we should load inactive banners, there's a query string specifically for that (`show_inactive ` just changed to `preview_unpublished`).

I think we're better off doing it like that otherwise we could run into a confusing bug where a dev assumes that time_context is tightly linked to showing inactive banners, and not realise that the behaviour is inferred.

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
EarthlingDavey: It's 0 if we're just 'Viewing' an active banner... seeing the banner the same as a logged out user.
Just `show_inactive` for looking at the Truss banner for example, no time travel necessary ;)
Both for the Sunak banner if we were 'Previewing' it last week.

web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
EarthlingDavey: Cheers, thanks for the suggestion to update it.

wilson1000 (approved):

PR #573, opened 2024-07-08
Localise times add add scheduled emails

EarthlingDavey:
This PR includes:

- Formatting of timestamps to local times
- Scheduled emails
- A fix for double ? in urls - including it here because it fixes a bug. Happy to change it if needs be.

The email content looks like:

image

feedback is welcome.


wilson1000 (approved):
Looks good 😊

PR #576, opened 2024-07-09
CDPT-1773 Delete events older than a year (but keep one).

EarthlingDavey:
This PR schedules a weekly event at 01:35

This runs an action to identify events older than 1 year and deletes them.

The most recent event is kept, so that we can always see who set the current toggle status.


wilson1000 (approved):
Nice :)

PR #582, opened 2024-07-12
CDPT-852 Staging deploy to match development

EarthlingDavey:


deploy/staging/secret.tpl.yml
```diff
@@ -36,3 +36,11 @@ type: Opaque
 data:
   AWS_CLOUDFRONT_PUBLIC_KEY: "${AWS_CLOUDFRONT_PUBLIC_KEY_BASE64}"
   # AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING: "${AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING_BASE64}"
```
wilson1000, line 38: Hey @EarthlingDavey 😄 
Is this needed, mate?

wilson1000 (approved):
Nice

deploy/staging/secret.tpl.yml
EarthlingDavey: Hey, I think it ill be helpful to keep the line `# AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING: "${AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING_BASE64}"` as it is part of a routine key rotation procedure.

I don't think I've documented it yet, but there should be something like a 5 step procedure to rotate the keys, with one of the steps being to uncomment this line.

PR #586, opened 2024-07-16
Prevent internal rewrites going to http and port 8080.

EarthlingDavey:


deploy/config/server.conf
```diff
@@ -24,6 +24,9 @@ server {
     root /var/www/html/public;
     index index.php;
 
+    # Prevents internal rewrites going to http and port 8080.
+    absolute_redirect off;
```
wilson1000, line 28: I might be wrong; internal redirection uses internal routing and never makes it out to the net. Internally, in the containers, HTTP is the only thing we can use.

Did I get that wrong? Is this only valid for requests being made that go out to the net and back again?

EarthlingDavey: It's one of those bugs that is pretty hard to spot because it doesn't happen that often.

It's linked to this PR https://github.com/ministryofjustice/justice-gov-uk/pull/203

Today I went to `https://dev.intranet.justice.gov.uk/wp/wp-admin` (without the slash) and it redirected me to `https://dev.intranet.justice.gov.uk:8080/wp/wp-admin/` (with port 8080).

This is because nginx is serving on port 8080 so it is adding that to the destination urls for redirects. To get around this we can turn off absolute redirects, so nginx returns relative redirects without the domain and protocol, and most importantly the port number.

PR #587, opened 2024-07-16
CDPT-1246 AS3CF increase batch size and add logging

EarthlingDavey:
This PR 

- adds filters to increase batch size for the AS3CF plugin.
- Adds a post instal/update script to that patches in logging to the plugin - similar to Justice UK patching technique.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -113,6 +113,8 @@ RUN make test
 
 FROM base-fpm AS build-fpm-composer
 
+WORKDIR /var/www/html
+
 ARG COMPOSER_USER
 ARG COMPOSER_PASS
 ARG AS3CF_PRO_USER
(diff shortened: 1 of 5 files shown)


Dockerfile
```diff
@@ -113,18 +113,20 @@ RUN make test
...
+
 ARG COMPOSER_USER
 ARG COMPOSER_PASS
 ARG AS3CF_PRO_USER
 ARG AS3CF_PRO_PASS
 
 COPY --from=composer:2 /usr/bin/composer /usr/bin/composer
 
-WORKDIR /var/www/html
+COPY ./bin/composer-auth.sh ./bin/composer-post-install.sh ./bin/
```
wilson1000, line 126: Can you pop `WORKDIR /var/www/html` on line 132 so `exec` can land us in the right place?

EarthlingDavey: Are you ok with it being a few lines up to match what we have on Justice UK? It's on 11 at the mo.

wilson1000: Mate, wherever you feel is best.

wilson1000: I didn't see it there lol

wilson1000 (approved):

Dockerfile
EarthlingDavey: haha, I was trying to simplify things, just made it more complicated 😅

PR #589, opened 2024-07-17
Allow frontend to cache and option to purge

wilson1000:


docker-compose.yml
```diff
@@ -197,5 +198,6 @@ services:
       target: build-cron
     environment:
       NGINX_HOST: ${NGINX_SERVICE_PORT}
+      NGINX_SERVICE_PORT: ${NGINX_SERVICE_PORT}
```
EarthlingDavey, line 201: Do we need to add `NGINX_SERVICE_PORT: "http://localhost:8080"` to `deploy/(demo|development|staging)/config.yml`

deploy/config/local/nginx/server.conf
```diff
@@ -75,16 +122,23 @@ server {
         deny all;
     }
 
+    # Deny access to load-scripts.php and load-styles.php to prevent DoS attacks.
```
EarthlingDavey, line 125: Nice, I think this block should also be in `deploy/config/server.conf`

EarthlingDavey (approved):
Built and runs locally 💯

docker-compose.yml
wilson1000: This was to solve a local problem where the NGINX_SERVICE_PORT (that is present in K8S) is available for the CRON to work

wilson1000: So basically, it is already in the environment 😄

EarthlingDavey:
See: https://github.com/ministryofjustice/intranet/blob/044aa597d8422b3f6c5747d6310d696811704d1f/docker/conf/nginx/server.conf#L15 

Having `$cookie_dw_agency` in the cache key will fix the issue. It will create individual caches for each agency.

Purging the cache might need a bit of work.

EarthlingDavey:
If we do:

`fastcgi_cache_key "$real_scheme$request_method$host$request_uri$cookie_dw_agency"; `

with $cookie_dw_agency at the end.

We can add a wildcard to the end of the purge line:

`fastcgi_cache_purge mojiCache "$real_scheme$request_method$host$1*";`

to purge a page for all agencies.

> If the [cache key](https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html#fastcgi_cache_key) of a purge request ends with an asterisk (“*”), all cache entries matching the wildcard key will be removed from the cache.

https://nginx.org/en/docs/http/ngx_http_fastcgi_module.html

EarthlingDavey:
While we're here, please could we also add info.php and errors.php onto the exclude list?

@wilson1000 if you're focusing on the other PR and (agree with my suggestions)... I don't mind adding these handful of changes.

EarthlingDavey:
There's one minor improvement that could be made.

The way I suggested to use `*` at the end of `fastcgi_cache_purge`, means that if:

1. We have cache for http://intranet.docker/guidance/ & http://intranet.docker/guidance/fire-health-safety/
2. We clear cache for guidance by hitting `http://intranet.docker/purge-cache/guidance/`

Then the cache for fire-health-safety will be purged, because the * not only matching the cookie value, but any path that's following `guidance`.

To address this, we can add a colon before the cookie value: e.g.
`fastcgi_cache_key ... $request_uri:$cookie_dw_agency";`
`fastcgi_cache_purge ... $host$1:*";`

I've tested this locally, and with the above scenario, `fire-health-safety` is not purged.

PR #591, opened 2024-07-17
CDPT-877 Add an ingress that is restricted by IP

EarthlingDavey:
This PR brings the IP level check out of the application and into Cloud Platform's ingress.

This way we can check an IP is allowed by sending a request to `https://dev-intranet-ip-restricted.apps.live.cloud-platform.service.justice.gov.uk/ip-restricted` (possibly by JS).

The feature is not complete, I will make application level changes in a different PR.

From the ticket:

- *.intranet.justice.gov.uk nor intranet.justice.gov.uk  is not routed through GlobalProtect.
- To save requesting Global Protect (and other VPNs?) to add *.intranet.justice.gov.uk to their routing we can have an auth endpoint at something like dev-intranet-api.apps.live.cloud-platform.service.justice.gov.uk
- *.service.justice.gov.uk  is already routed via Global Protect.
- This will be similar to the existing setup - it will not route all intranet browsing traffic through Global Protect, this is a good thing for saving bandwidth and low latency.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -47,6 +47,7 @@ jobs:
           # AWS_CLOUDFRONT_PUBLIC_KEY_EXPIRING: "${{ secrets.AWS_CLOUDFRONT_PUBLIC_KEY_B }}"
           BASIC_AUTH_USER: ${{ secrets.BASIC_AUTH_USER }}
           BASIC_AUTH_PASS: ${{ secrets.BASIC_AUTH_PASS }}
+          ALLOWED_IPS: ${{ secrets.ALLOWED_IPS }}
         run: |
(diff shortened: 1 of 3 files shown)


wilson1000:
Hey buddy, I can't get my head around the architecture of what this does. Can we have a chat about it, just to make sure it's maintainable and will work as we expect?

wilson1000:
Also, we won't be routing users through Global Protect, they will need to log in through Azure or we need to allow their IP range. I think this is already accepted but not sure why it was mentioned. I'd love some clarity 😄  cheers @EarthlingDavey

EarthlingDavey:
Closed as we will likely pull in allowed IPs from a private repo. And, we do not need to do an ingress level ip allowed list.

PR #596, opened 2024-07-19
CDPT-1773 Create manage_prior_party_banners capability and conditionally hide metabox

EarthlingDavey:


web/app/themes/clarity/inc/admin/prior-party/prior-party-banner-admin.php
```diff
@@ -93,6 +104,11 @@ public function __construct()
         // Add the delete action to the schedule.
         add_action('prior_party_banner_event_cleanup_cron_hook', [$this, 'deleteOldTrackEvents']);
 
+        // Ensure administrator always has the required capability.
+        if (current_user_can('administrator')) {
+            $this->required_capability = 'administrator';
+        }
```
EmilyHazlehurst, line 110: @EarthlingDavey Is this necessary? Should `private string $required_capability = 'manage_prior_party_banners';` not just be set to 'administrator' instead?

EmilyHazlehurst: Or are there cases where non-admin users are give the prior_party_banners_capability?

EmilyHazlehurst: Oh, is there a difference between administrator and agency_admin? In which case I've answered my own question 😅

EmilyHazlehurst (approved):
Looks good :)

PR #598, opened 2024-07-22
CDPT-1760 - Pull in IP ranges from maintained repository.

EarthlingDavey:


wilson1000:
Hey, this looks great @EarthlingDavey  😄 
Just a minor; `ips.yml` isn't intuitive, could this be renamed to `ip-ranges-configure.yml` or something similar?

EarthlingDavey:
Thanks @wilson1000 , I've renamed to `get-ip-ranges.yml` and renamed the job to `get_ip_ranges`.

This means that lines like the following will read nicely. `${{ needs.get_ip_ranges.outputs.ip_ranges }}`

What do you think?

wilson1000:
> ip-ranges-configure.yml

If we think of this like a one-method class, you could name the class `ip-ranges-configure.yml` with a job node called get-ip-ranges` this makes really good sense when reading for the first time.

Does that sound ok?

wilson1000:
> This means that lines like the following will read nicely. `${{ needs.get_ip_ranges.outputs.ip_ranges }}`

With the example above could we get these lines to read like this instead:

`${{ needs.ip_ranges.outputs.get_ip_ranges }}`

PR #611, opened 2024-07-24
CDPT-1864 Add export CSV filter function and notes

EarthlingDavey:


public/app/themes/clarity/inc/admin/plugins/wp-all-export.php
```diff
@@ -0,0 +1,84 @@
...
+     * 
+     * @param array $articles the rows being exported.
+     * 
+     * @return array the filtered and mapped rows.
+     */
+
+    public function customOperationsAreaFilter($articles)
+    {
+
+        $parent_id = 69769;
```
wilson1000, line 42: There's a hard-coded number here, is this expected?

EarthlingDavey: Good spot :) I've replaced with an env var, and updated the instructions.

wilson1000 (approved):

PR #612, opened 2024-07-26
CDPT-887 auth with nginx cache

EarthlingDavey:
This PR:

### IPs

- Moves the allowed IP matching from app level to nginx.
- Utilises the nginx_geo module for assigning a value to a matching IP ranange. i.e. 1 for allow and 2 for deprecated.
- Uses an nginx- init script to populate a config file from the `IPS_FORMATTED` env var.
- A header `HTTP_X_IP_GROUP` is available to our application incase we want to show a deprecation warning. If we want to do this, we may need to add $geo to the `fastcgi_cache_key` - or fetch the deprecation warning via an un-cached url like ajax.php

### Oauth

- Creates an `/auth/verify` endpoint that returns a 200 / 401 status, depending on a valid JWT. This endpoint doesn't load WordPress.
- Creates a dynamic 401 page that will redirect to login, on 5 failed attempts the 401 page is shown.

### Caching

Care has been taken to only serve cached results if the user is allowed.
Previously cached hits were served to non-auth'd users, this has been fixed.


public/app/mu-plugins/moj-auth/moj-auth.php
```diff
@@ -12,19 +12,20 @@
...
 use Roots\WPConfig\Config;
 
 // Do not allow access outside WP
 defined('ABSPATH') || exit;
 
 // If the plugin isn't enabled, return early.
-if(Config::get('MOJ_AUTH_ENABLED') === false) {
+if (Config::get('MOJ_AUTH_ENABLED') === false) {
     return;
 }
```
EarthlingDavey, line 24: Double check return works here.

deploy/config/local/nginx/server.conf
```diff
@@ -93,22 +104,73 @@ server {
...
         try_files $uri $uri/ /index.php?$args;
     }
 
+    # @see https://gock.net/blog/2020/nginx-subrequest-authentication-server
+    location = /auth/verify {
+        # Internal only, so /auth/verify can not be accessed from outside.
+        internal;
+
+        if ( $geo != 0 ) {
+            add_header X-Geo $geo;
```
EarthlingDavey, line 148: TODO namespace this to make cleat that we've added it.

public/app/mu-plugins/moj-auth/401.php
```diff
@@ -0,0 +1,98 @@
...
+    use AuthJwt;
+    use AuthUtils;
+
+    private $now            = null;
+    private $debug          = false;
+    private $https          = false;
+    private $sub            = '';
+
+    const OAUTH_LOGIN_URI = '/auth/login';
+    const MAX_AUTO_LOGIN_ATTEMPTS = 5;
```
EarthlingDavey, line 42: This might need to be 1. As we're expecting Entra to work.

public/app/mu-plugins/moj-auth/traits/jwt.php
```diff
@@ -2,8 +2,8 @@
 
 namespace MOJ\Intranet;
 
-// Do not allow access outside WP
-defined('ABSPATH') || exit;
+// Do not allow access outside WP, 401.php or verify.php
+defined('ABSPATH') || defined('DOING_STANDALONE_401') || defined('DOING_STANDALONE_VERIFY') || exit;
```
EarthlingDavey, line 6: Add a phpdoc block with @uses and link to the other files.

public/app/mu-plugins/moj-auth/templates/401-redirect.php
```diff
@@ -0,0 +1,32 @@
...
+        <meta http-equiv="refresh" content="0; url=<?= $this::OAUTH_LOGIN_URI ?>">
+        <script type="text/javascript">
+            window.location.href = "<?= $this::OAUTH_LOGIN_URI ?>"
+        </script>
+        <title>Page Redirection</title>
+    </head>
+
+    <body>
+        <!-- Note: don't tell people to `click` the link, just tell them that it is a link. -->
+        If you are not redirected automatically, follow this <a href='<?= $this::OAUTH_LOGIN_URI ?>'>link to login</a>.
```
EarthlingDavey, line 29: Click here to login.

deploy/config/local/nginx/server.conf
```diff
@@ -93,22 +104,73 @@ server {
...
+        # Internal only, so /auth/verify can not be accessed from outside.
+        internal;
+
+        # The 401 handler, WordPress is not loaded in this file.
+        set $script_name /app/mu-plugins/moj-auth/$1.php;
+        include /etc/nginx/php-fpm-auth.conf;
+    }
+
+    # Rewrite auth endpoints to fpm (WordPress's index.php)
+    location ~ ^/auth/(login|callback|heartbeat|logout) {
```
EarthlingDavey, line 168: Remove `heartbeat|logout` double check we don't need a block.

public/app/themes/clarity/src/globals/js/auth-heartbeat.js
```diff
@@ -0,0 +1,6 @@
+export default (function ($) {
+    // Sent a request to the heartbeat endpoint, this will refresh the oauth token.
+    setInterval(function(){
+        $.get( "/auth/heartbeat" )
```
EarthlingDavey, line 4: If 401, modal - your session has ended, please login again for security purposes. Blur bg. Redirect in 5s. 

The js can just reload the current page.

Change to 10s

deploy/config/local/nginx/server.conf
```diff
@@ -93,22 +104,73 @@ server {
...
     }
 
+    # Don't cache any auth paths.
+    if ($request_uri ~* "/auth/") {
+        set $skip_cache 1;
+    }
+
     # ...it's from a logged in user, the cookie 'wordpress_no_cache' exists, or we're mid moj-auth flow.
     if ($http_cookie ~* "comment_author|wordpress_[a-f0-9]+|wp-postpass|wordpress_no_cache|wordpress_logged_in|OAUTH_.*") {
         set $skip_cache 1;
```
EarthlingDavey, line 114: Remove `|OAUTH_.*`

deploy/config/local/nginx/server.conf
```diff
@@ -93,22 +104,73 @@ server {
...
     }
 
+    # Don't cache any auth paths.
+    if ($request_uri ~* "/auth/") {
```
EarthlingDavey, line 108: Add `^`

deploy/config/local/nginx/server.conf
```diff
@@ -26,6 +26,12 @@ fastcgi_cache_use_stale updating error timeout invalid_header http_500;
 fastcgi_cache_key "$request_method$host$request_uri:$cookie_dw_agency";
 fastcgi_ignore_headers Cache-Control Expires Set-Cookie;
 
+
+geo $geo {
```
EarthlingDavey, line 30: Rename `$geo`

deploy/config/local/nginx/server.conf
```diff
@@ -93,22 +104,73 @@ server {
...
+    if ($request_uri ~* "/auth/") {
+        set $skip_cache 1;
+    }
+
     # ...it's from a logged in user, the cookie 'wordpress_no_cache' exists, or we're mid moj-auth flow.
     if ($http_cookie ~* "comment_author|wordpress_[a-f0-9]+|wp-postpass|wordpress_no_cache|wordpress_logged_in|OAUTH_.*") {
         set $skip_cache 1;
     }
 
+    # Skip cache if the concatenated value doesn't contain 200 - the user is not allowed.
```
EarthlingDavey, line 117: Use xincatenation and `!~` regex because we cant use the OR operator in nginx

public/app/mu-plugins/moj-auth/moj-auth.php
EarthlingDavey: > If called from the global scope, then execution of the current script file is ended. If the current script file was included or required, then control is passed back to the calling file.

Looking ok.

public/app/mu-plugins/moj-auth/401.php
EarthlingDavey: Set to 1. And set the retry attempts to be logged in a session cookie. User can reset the count to 0 by restating the browser.

PR #613, opened 2024-07-29
CDPT-887 amends based on previous PR and workflow update.

EarthlingDavey:


public/app/mu-plugins/moj-auth/401.php
```diff
@@ -72,17 +77,20 @@ public function handle401Request(): void
         // Set loginAttempts with a default of 1, or add one to the existing value.
         $jwt->login_attempts = empty($jwt->login_attempts) ? 1 : ((int) $jwt->login_attempts) + 1;
 
-        // Where to redirect the user after successful login.
-        $jwt->success_uri = $_SERVER['REQUEST_URI'];
+        // Always add the schema and domain here, to prevent an open redirect vulnerability.
+        $jwt->success_url = $_ENV['WP_HOME'] . $_SERVER['REQUEST_URI'];
```
EarthlingDavey, line 81: @wilson1000 I updated this as our previous solution used a WP function, get_home, which isn't available to us here.

See also the above check for `if(empty($_ENV['WP_HOME']))`

deploy/config/local/nginx/server.conf
```diff
@@ -79,7 +79,7 @@ server {
     }
 
     # We use the header X-Ip-Group internally, if the client sends this header return Bad Request.
```
wilson1000, line 81: Should the comment be updated?

public/app/mu-plugins/moj-auth/401.php
```diff
@@ -54,11 +54,16 @@ public function __construct(array $args = [])
             http_response_code(401) && exit();
         }
 
-        if (!file_exists($this::REDIRECT_TEMPLATE)) {
+        if (!file_exists($this::STATIC_401_REDIRECT)) {
             error_log('moj-auth/401.php template was not found.');
```
wilson1000, line 58: Should this be...
```php
error_log('moj-auth/401.php ' . basename($this::STATIC_401_REDIRECT) . ' template was not found.');
```

wilson1000 (approved):

public/app/mu-plugins/moj-auth/401.php
EarthlingDavey: Good point, thanks :)

wilson1000 (approved):

PR #617, opened 2024-07-30
Use MAX_FAILED_CALLBACKS instead of MAX_AUTO_LOGIN_ATTEMPTS

EarthlingDavey:


public/app/mu-plugins/moj-auth/401.php
```diff
@@ -71,23 +71,33 @@ public function handle401Request(): void
...
+        $this->log('Request URI: ' . $_SERVER['REQUEST_URI']);
 
         // Always add the schema and domain here, to prevent an open redirect vulnerability.
         $jwt->success_url = $_ENV['WP_HOME'] . $_SERVER['REQUEST_URI'];
 
         // Set the cookie expiry to 0 to create a session cookie.
         $jwt->cookie_expiry = 0;
 
+        // Set failed_callbacks with a default of 0.
+        $jwt->failed_callbacks = isset($jwt->failed_callbacks) ? $jwt->failed_callbacks : 0;
```
wilson1000, line 92: This might be better, null coalesce:

```php
$jwt->failed_callbacks = $jwt->failed_callbacks ?? 0;
```

Where is `$failed_callbacks` being declared, is it in a trait or class? 
It looks like the var is being dynamically set.

public/app/mu-plugins/moj-auth/moj-auth.php
```diff
@@ -120,6 +120,19 @@ public function handleCallbackRequest(): void
...
             $this->log('Access token is not valid, or expired.');
+
+            // Update (or create) the JWT to keep track of failed callbacks.
+            $jwt = $this->getJwt() ?: (object)[];
+
+            // Set to 0 for a session cookie.
+            $jwt->cookie_expiry = 0;
+
+            // Set failed_callbacks with a default of 1, or add one to the existing value.
+            $jwt->failed_callbacks = isset($jwt->failed_callbacks) ? ((int) $jwt->failed_callbacks) + 1 : 1;
```
wilson1000, line 131: Assuming this isn't being dynamically created, If the variable is NULL, `isset()` isn't needed.
```php
$jwt->failed_callbacks = $jwt->failed_callbacks ? ((int) $jwt->failed_callbacks) + 1 : 1;
```

public/app/mu-plugins/moj-auth/traits/jwt.php
```diff
@@ -105,14 +105,14 @@ public function setJwt(object $args = new \stdClass()): object
...
+
+        // Custom claims - conditionally add failed_callbacks from $args or class property.
+        if (isset($args->failed_callbacks)) {
+            $payload['failed_callbacks'] = $args->failed_callbacks;
         }
-        
+
         // Custom claims - conditionally add success_url from $args or class property.
-        if(!empty($args->success_url)) {
+        if (!empty($args->success_url)) {
```
wilson1000, line 115: If `$args->success_url` is NULL, do we need `empty()`?

wilson1000 (approved):
Cool mate, just some comments

public/app/mu-plugins/moj-auth/traits/jwt.php
EarthlingDavey: We still need it because sometimes `setJwt` will be called without it being set.

public/app/mu-plugins/moj-auth/moj-auth.php
EarthlingDavey: I think there will be edge cases where it's not set. It should never be null, if null is passed to `setJwt` then the property `failed_callbacks` is not there.

public/app/mu-plugins/moj-auth/401.php
EarthlingDavey: Good shout.

On the usual login flow, this will be the first place that `$jwt->failed_callbacks` is set.

EarthlingDavey:
@wilson1000  Thanks for the comments, I've addressed with a couple of explanations :)

public/app/mu-plugins/moj-auth/traits/jwt.php
wilson1000: https://www.php.net/manual/it/language.oop5.properties.php#language.oop5.properties.dynamic-properties
> Dynamic properties are deprecated as of PHP 8.2.0. It is recommended to declare the property instead.

EarthlingDavey: Hey @wilson1000 , thanks for the link. I've not ignored you, I replied in Slack :) https://mojdt.slack.com/archives/C0744UHUY3X/p1722346954304479

PR #626, opened 2024-07-31
Backdrop modal heartbeat UAT

wilson1000:


public/app/themes/clarity/src/globals/js/auth-heartbeat.js
```diff
@@ -1,6 +1,154 @@
...
+        },
+        feedback: () => {
+            // We can use this to place a box in the bottom right corner of
+            // the screen. This box can remind the user they are UATing and
+            // offer a link to submit feedback.
+            console.log('Display the feedback box.');
+        },
+        confirm: () => {
+            const title = 'You are about to view a test version of the MoJ Intranet'
+            const html = '<br />If you intend to assist in user acceptance testing, please continue.<br /><br />' +
```
EarthlingDavey, line 105: Just a suggestion, but could you use the word participate instead.

e.g. To participate in user acceptance testing, please continue.

public/app/themes/clarity/src/globals/js/auth-heartbeat.js
```diff
@@ -1,6 +1,154 @@
...
+              '<button class="modal-continue primary" type="button">&nbsp; Continue &nbsp;</button> &nbsp; or &nbsp; ' +
+              '<button class="modal-escape primary" type="button">&nbsp; Visit the live Intranet &nbsp;</button>'
```
EarthlingDavey, line 124: A minor consideration, could this be an a with `a` tag with `href`.

public/app/themes/clarity/src/globals/js/auth-heartbeat.js
```diff
@@ -1,6 +1,154 @@
...
+                for (ii; ii < runOn.length; ii++) {
+                    if (location.href.includes(runOn[ii])) {
+                        return true;
+                    }
+                }
+                return false;
+            }
+        },
+        modal: (title, html) => {
+            let modal = $('<div\>', { 'class': 'heartbeat__modal' })
```
EarthlingDavey, line 99: If you wanted to use semantic elements, `<dialog>` might work here.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog

EarthlingDavey (approved):
This looks really nice. I've added a couple of minor comments.

My last thought is that the UAT testing modal is not really related to the auth heartbeat.

Given all the time in the would it would probably be best practice to:

- abstract the modal code away to a component
- have UAT js code in a separate file
- not enqueue the UAT .js on production

I'll leave that up to your consideration, as there are various factors to consider e.g. launch timelines 😁

public/app/themes/clarity/src/globals/js/auth-heartbeat.js
```diff
@@ -1,6 +1,188 @@
...
+             *
+             * @returns {boolean}
+             */
+            can_run: () => {
+                const runOn = [
+                    'dev.intranet',
+                    'intranet.docker',
+                    'staging.intranet'
+                ];
+                let ii = 0;
```
EarthlingDavey, line 86: Could this be simplified with using array `includes` and `location.hostname`.

wilson1000:
> Given all the time in the world it would probably be best practice to:
> 
> * abstract the modal code away to a component
> * have UAT js code in a separate file
> * not enqueue the UAT .js on production

I'd considered that already 😄 we don't have all the time in the world at the moment and wanted to get something out that was working in the way you created our heartbeat.

One other consideration is; that we will need to make sure the heartbeat code is bundled with the MU plugin. It's currently installed in the theme. As you mentioned we should make sure that relevant code is kept together, or broken out into a logical format.

PR #629, opened 2024-08-01
Update node containers and theme packages

EarthlingDavey:
- Update node containers to `22-alpine`.
- As alpine doesn't have `make`, move install and run scripts to `docker-compose`
- Update outdated dependencies and remove `vue-template-compiler`.
- Run `npm audit` and `fix`.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,10 +59,9 @@ services:
 
   node:
     container_name: intranet-node
-    image: node:20
+    image: node:22-alpine
     working_dir: /node
-    entrypoint: make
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -59,10 +59,9 @@ services:
 
   node:
     container_name: intranet-node
-    image: node:20
+    image: node:22-alpine
     working_dir: /node
-    entrypoint: make
```
wilson1000, line 64: If we are removing Make, let's make sure we remove from volume mounts too.

PR #634, opened 2024-08-05
Notification for UAT

wilson1000:
This temporary feature utilises the heartbeat JS architecture to present a notification and a link to the UAT user.

## Notification
Screenshot 2024-08-05 at 11 51 31

## Microsoft form
Screenshot 2024-08-05 at 11 28 54


EarthlingDavey (approved):
Looks all good. Just a one of question and a comment.

Do we want the feedback button to be on admin views?

We could consider using [sentry feedback](https://docs.sentry.io/platforms/javascript/user-feedback/) in the future, that way we'll get some additional info about the client's session.

wilson1000:
Hey @EarthlingDavey, new updates.

Just a note (and you probably know this) - heartbeat doesn't work in the admin screens. Is there any reason for this?

EarthlingDavey:
Looks really good!

> heartbeat doesn't work in the admin screens. Is there any reason for this?

Good catch! This wasn't an intentional choice - I hadn't noticed or thought about it.

I think it probably should be on admins screens, do you agree?

We would have to be careful that editors don't lose unsaved work.

wilson1000:
> We would have to be careful that editors don't lose unsaved work.

They have the WP revision system underway whilst they are working. This saves the work in their browser. We should work out a way to test it on the admin side, my concern is the same as yours.

PR #644, opened 2024-08-07
CDPT-1909 - fixes styled dropdowns.

EarthlingDavey:
- Pulls in latest ACF from the official repo.
- Adds ACF key during build and runtime.
- Fixes styled dropdowns.


bin/composer-auth.sh
```diff
@@ -20,8 +20,9 @@ then
 	cat <<- EOF >> auth.json
 		{
 			"http-basic": {
-			  "composer.wp.dsd.io": {"username": "$COMPOSER_USER","password": "$COMPOSER_PASS"},
-        "composer.deliciousbrains.com": {"username": "$AS3CF_PRO_USER","password": "$AS3CF_PRO_PASS"}
+				"composer.wp.dsd.io": {"username": "$COMPOSER_USER","password": "$COMPOSER_PASS"},
+				"composer.deliciousbrains.com": {"username": "$AS3CF_PRO_USER","password": "$AS3CF_PRO_PASS"},
+				"connect.advancedcustomfields.com": {"username": "$ACF_PRO_LICENSE", "password": "https://intranet.justice.gov.uk"}
```
wilson1000, line 25: Hi @EarthlingDavey - Did you set up basic authentication using our domain as a password?

wilson1000 (approved):
Just one comment my friend. Thanks for this.

bin/composer-auth.sh
wilson1000: Our security department will flag this as the 'password' has a value. We might want to stop them from messaging us by popping the value in a hidden variable.

EarthlingDavey: Hey, good spot. I didn't set the password, it's how acf does the auth.

https://www.advancedcustomfields.com/resources/installing-acf-pro-with-composer/#build-systems-and-environment-variables

Do you think commenting would be sufficient, or set an environment variable?

PR #652, opened 2024-08-08
CDPT-1903 Auth heartbeat for admin

EarthlingDavey:
The idea is to have a slightly different behaviour on admin screens.

Instead of reloading, the user will open a new tab for the auth flow.
This will prevent any work or state like scroll position form being lost.
After the separate tab has completed the login flow, it will close - or say you may now close this tab.
Then the user will return to the admin screen, to see a loading ~~spinner~~ label for < 10s.
And then back to where they left off.

To test locally:

- remove IP from IPS_FORMATTED
- login to an admin screen.
- delete only the jwt cookie.
- follow the Login in new tab button.


wilson1000 (approved):
Looks good 😄

PR #660, opened 2024-08-12
Fix attempt - modify internal verify endpoint for POSTs.

EarthlingDavey:


deploy/config/php-fpm-auth.conf
```diff
@@ -8,5 +8,9 @@ include fastcgi_params;
 fastcgi_param SCRIPT_NAME $script_name;
 fastcgi_pass fpm;
 
+# Force the request method to get. Otherwise, when verifying POSTs, fpm will be waiting for a non-existing payload.
+# See: https://forum.nginx.org/read.php?2,257968
+fastcgi_param REQUEST_METHOD "GET";
```
wilson1000, line 13: We need to isolate the setting to the `upload.php` script. Sadly, this approach is too broad.

wilson1000 (changes requested):
Can we isolate the setting to the `upload.php` path, please?

deploy/config/php-fpm-auth.conf
wilson1000: Haha, read the file name 🤦‍♂️ 
`deploy/config/php-fpm-auth.conf`

wilson1000 (approved):

PR #663, opened 2024-08-13
Remove unused fields

wilson1000:


EarthlingDavey:
Looks good, though I can't fully test locally, because I can't trigger the issue.

PR #669, opened 2024-08-14
Custom fpm metrics. Add metric exporter to local and deployment.

EarthlingDavey:
This is giving us metrics locally.

image

Lets test if these are working on dev. Then I'll try and set the prometheus rules.


EarthlingDavey:
Having swapped out the 3rd party exporter, we can still access the following metrics.

image

wilson1000 (commented):
The endpoint should be `/metrics` for the CP Prometheus service to automatically pick it up.

PR #682, opened 2024-08-28
Manage versions in Dockerfile 

wilson1000:


EarthlingDavey:
I'm not sure if this will work for nginx.

Because sometimes because nginx & nginxinc/nginx-unprivileged will have different patch versions available for 1.26

They should match, otherwise a a later step in the Dockerfile will throw an error.

EarthlingDavey (approved):
It looks like it's working.

There's a chance we might need to re-visit if nginx and  nginxinc/nginx-unprivileged release new 1.26 patch versions at different times.

PR #683, opened 2024-08-29
Add newlines to ips_formatted

EarthlingDavey:
Post deploy, check with:

```
NSP=intranet-dev 
kubectl -n $NSP get secrets intranet-dev-base64-secrets -o json | jq -r  ".data.IPS_FORMATTED" | base64 -d
```

If there are newlines between ips then this PR has worked.


wilson1000 (approved):
Looks good @EarthlingDavey 😄

PR #689, opened 2024-09-09
CDPT-1239 - Update nginx config and add service & pod monitors.

EarthlingDavey:
This PR:

- Adds `$ip_skip_oauth`, this allows us to block `Cloud Platform IPs` from accessing the service.

  This in turn means the `metrics/service` endpoint will correctly report a status code of 401 when trying to access the homepage on staging & production.

- ServiceMonitors and PodMonitors have been declared for Demo and Staging environments.

- Alerts have been declared for dev, after this PR, similar alerts will be set for demo and staging.


wilson1000 (approved):
This looks good @EarthlingDavey

PR #691, opened 2024-09-09
CDPT-1239 Correctly allow access to `/metrics/service`

EarthlingDavey:
Re use of 404s here.

I've used them, else it redirects the user to login.

> Servers may also send this response instead of 403 Forbidden to hide the existence of a resource from an unauthorized client. This response code is probably the most well known due to its frequent occurrence on the web.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#client_error_responses

### DIFF
--- a/deploy/config/server.conf
+++ b/deploy/config/server.conf
@@ -282,10 +282,9 @@ server {
         fastcgi_pass fpm;
     }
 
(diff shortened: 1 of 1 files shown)


wilson1000 (approved):
Nice

PR #698, opened 2024-09-12
Ability to turn Sentry on or off

wilson1000:
There is a noticable lag on the staging stack, even with the same resources used. 

This update is to help us turn Sentry off to see if it is causing latency.

### DIFF
--- a/public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
+++ b/public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
@@ -4,6 +4,9 @@ namespace DeliciousBrains\WP_Offload_Media\Tweaks;
 
 use Exception;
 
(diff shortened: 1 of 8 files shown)


public/app/themes/clarity/inc/amazon-s3-and-cloudfront-signing.php
```diff
@@ -246,7 +258,7 @@ public function getSignedCookie(): array
      * Timeline
      * 1. Does the user have a signed cookie with a long expiry in their browser?
      *   a. Yes: Do nothing.
-     *   b. No: Continue.
+     *   B. No: Continue.
```
EarthlingDavey, line 261: There's a mis-match of capitalisation here.

EarthlingDavey (approved):
Looks good 😊

PR #699, opened 2024-09-12
Decouple metrics/service from WP - so that it works when DB is down.

EarthlingDavey:


wilson1000:
Hey @EarthlingDavey, there was an error with the Docker build, I fixed it but have a query about this PR...
Do we need to allow access to metrics if the DB is down? IMHO, if the DB isn't attached, the app is useless, meaning we won't have to worry about collecting metrics - what do you think?

EarthlingDavey:
Thanks for fixing the build :)

My initial thinking with this was that it should fix the alerts in #cdpt-intranet-and-justice-cloud-platform-alerts , as the /metrics/service endpoint is a 500 error from 11pm to 7am.

Then, I realised it is probably a good idea to be as decoupled (as practically possible) from WordPress. The DB going down is only one way that the /metrics/service endpoint could go down. If we unfortunately mess up some env vars and WordPress is erroring, it would be nice to still be assured that the site is secure.

I think money/time no object we should completely decouple this monitoring/alerting from the cluster. But, considering the timeline decoupling from the application code might be an ok compromise.

wilson1000 (approved):
Nice. Let's see how this flies 👍

PR #700, opened 2024-09-17
Optimise database calls

EarthlingDavey:
This PR

- Moves role updates to an Admin Action. Meaning we avoid writing to the database 10 times during front end page views.
- Cache heavy database query for c-left-hand-menu. A function that was slow and used 50mb of memory, is now cached for 5mins. It speeds up page loads for pages like HR.
- Add env var `MOJ_AUTH_DEBUG` to clear up php logs.


config/application.php
```diff
@@ -170,7 +170,7 @@
 Config::define('SCRIPT_DEBUG', false);
 ini_set('display_errors', '0');
 // Additional logging for the authentication mu-plugin.
-Config::define('MOJ_AUTH_DEBUG', true);
+Config::define('MOJ_AUTH_DEBUG', env('MOJ_AUTH_DEBUG') ?? false);
```
wilson1000, line 173: Hi @EarthlingDavey 
Just an observation; why are we setting a second global variable? 

This...
```php
echo env('MOJ_AUTH_DEBUG');
echo MOJ_AUTH_DEBUG;
```
It will yield the same value.

wilson1000: Oh... I see. The constant can be false as well as true. Is that right?

public/app/themes/clarity/inc/admin/users/add-agency-admin.php
```diff
@@ -8,6 +8,10 @@
  * Access to everything within that agency minus a few Administrator settings.
  *
  * Type of user: heads of agencies, non-developer intranet team members
+ * 
+ * If you edit this file, you must sync the roles to the database by running the 
+ * SyncUserRoles admin action. 
+ * Navigate to Tools > Admin Commands > Sync user roles from codebase to database
```
wilson1000, line 14: Can we create a shell script to execute when the container starts for the first time, instead of relying on a manual action?

```shell
wp run my-custom-command
```

wilson1000 (commented):
Hey mate, sorry I've been out of action today. 
I've made some comments around automating the user sync process. 
Basically, as we are using Docker and we resync users only when a file changes, can we run a command to sync each time the container starts using the init script process?

config/application.php
EarthlingDavey: Ah good spot! Considering I'm only checking if it's true in the codebase, I don't think I need to set the default value as false.

public/app/themes/clarity/inc/admin/users/add-agency-admin.php
EarthlingDavey: Great shout 🙌

wilson1000 (approved):
Looks great @EarthlingDavey

EarthlingDavey:
Thanks @wilson1000 🙌

EarthlingDavey:
Te be merged after this wordpress base image build... 

https://github.com/ministryofjustice/wordpress-base-fpm/actions/runs/10925980980

PR #701, opened 2024-09-18
Fix mismatch between localhost and Cloud Platform internal IPs.

EarthlingDavey:
This PR fixes the regression of ticket: https://dsdmoj.atlassian.net/jira/software/c/projects/CDPT/boards/1154?issueParent=224477%2C314124&selectedIssue=CDPT-1928


wilson1000 (approved):
Nice 😄

PR #704, opened 2024-09-18
Get redis env vars from secrets and update default config.

EarthlingDavey:


config/application.php
```diff
@@ -181,18 +181,18 @@
...
+Config::define('WP_REDIS_CLIENT', env('WP_REDIS_CLIENT') ?: 'relay');
+// The password of the Redis server.
+Config::define('WP_REDIS_PASSWORD', env('WP_REDIS_PASSWORD'));
 
 // Advanced config...
 
-// Disable the caching via env var - in case of emergency.
-// Config::define('WP_REDIS_DISABLED', true);
+// Disable the caching if WP_REDIS_HOST or via WP_REDIS_DISABLED - in case of emergency.
+Config::define('WP_REDIS_DISABLED', empty(Config::get('WP_REDIS_HOST')) || env('WP_REDIS_DISABLED'));
```
wilson1000, line 195: At this point, `Config` [hasn't been applied](https://github.com/ministryofjustice/intranet/blob/ada4ff6617d1d5374474b5e07c58eb34c2a167be/config/application.php#L233).

Have we tested if `Config::get` returns the expected value?

wilson1000 (approved):
Looks good @EarthlingDavey - I've left some comments for consideration, nothing more 😄

config/application.php
EarthlingDavey: Good shout, I did check yesterday... double checked and it works as expected.

I think the apply is what solidifies the config into constants, but `Config::get` can be used before `Config::apply`.

Side-note, I have just realised that throughout the theme I've used `Config::get` in various places.

e.g. I could replace 

```
if (Config::get('MOJ_AUTH_ENABLED') === false) {
    return;
}
```

with 

```
if (MOJ_AUTH_ENABLED === false) {
    return;
}
```

It makes sense right?

config/application.php
```diff
@@ -181,18 +181,18 @@
...
-// The prefix used for all cache keys to avoid data collisions, should be human readable
-Config::define('WP_REDIS_PREFIX', env('WP_REDIS_PREFIX') ?: 'local');
+Config::define('WP_REDIS_SCHEME', env('WP_REDIS_SCHEME') ?: 'tls');
+// The prefix used for all cache keys to avoid data collisions, should be human readable.
+Config::define('WP_REDIS_PREFIX', env('WP_REDIS_PREFIX') ?: WP_ENV);
 // Relay is the fastest redis client.
-Config::define('WP_REDIS_CLIENT', env('WP_REDIS_CLIENT') ?: 'predis'); 
-
-// Config::define('WP_REDIS_PASSWORD', env('WP_REDIS_PASSWORD')); 
+Config::define('WP_REDIS_CLIENT', env('WP_REDIS_CLIENT') ?: 'relay');
```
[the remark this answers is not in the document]
EarthlingDavey: The answer to this and your question about `WP_REDIS_PREFIX ` is ... it depends 😁

During stress testing we could trial out the 3 options `predis|phpredis|relay`.

On dev we can exec into the pod and make live changes to work out which config. is optimal.

On staging, because we have multiple pods, the quickest way to change the setting would be:

```
kubectl edit configmap intranet-staging -n $NSP
kubectl rollout restart deployment/intranet-staging -n $NSP
```

I agree, that after stress testing we should remove these :)

config/application.php
wilson1000: Yes, it makes good sense. Double check we can still access them though 😄

PR #705, opened 2024-09-19
CDPT-1239 Update dashboard and alerts to cater for basic auth on dev & demo.

EarthlingDavey:
This PR removes 2 unnecessary alerts and updates the dashboard.

This is because, when basic auth is enabled (on dev and demo), there is no way to test for a 400 status from the nginx pod.


wilson1000 (approved):
Looks good dude 😄

PR #706, opened 2024-09-19
Update caching plugin, and dev. & staging deployments.

EarthlingDavey:
Having done some live testing on the dev. environment, this PR includes the optimal config.

It also updates the development deployment, and adds CACHE env vars to the staging deployment.


composer.json
```diff
@@ -145,10 +145,10 @@
             "@php -r \"file_exists('.env') || copy('.env.example', '.env');\""
         ],
         "test": [
-            "vendor/bin/phpcs -d memory_limit=256M"
+            "vendor/bin/phpcs -d memory_limit=256M --ignore=public/app/object-cache.php"
```
EmilyHazlehurst, line 148: @EarthlingDavey would it not still be useful to test this? Or is there an overwhelming number of formatting issues since it's from a separate codebase?

Since this won't be treated like a plugin then we won't be alerted to any updates which could mean we miss warnings about deprecated functions, for example.

EmilyHazlehurst: Having said that, it doesn't look like we're checking that at the moment so maybe something to think about in the future :)

EmilyHazlehurst (approved):

composer.json
EarthlingDavey: Thanks for flagging this. The test was on mostly [formatting errors](https://github.com/ministryofjustice/intranet/actions/runs/10957778490/job/30426616954), and if I formatted the file according to the project's standards it would make it really tough to diff agains upstream changes.

I'll keep it like this for now, if we add more through tests to our suite then we can reconsider ignoring it.

PR #713, opened 2024-09-27
CDPT-1896 Change the local and dev Entra apps to the Entra dev environment.

EarthlingDavey:


.github/README.md
```diff
@@ -333,15 +331,15 @@ To verify that S3 & CloudFront are working correctly.
...
-   - Redirect URI: `Web` and `http://localhost/oauth2?action=callback`  
-     or `https://dev.intranet.justice.gov.uk/oauth2?action=callback` etc.
+   - Redirect URI: `Web` and `https://staging.intranet.justice.gov.uk/oauth2/callback`  
+     or `https://intranet.justice.gov.uk/oauth2/callback` etc.
 6. Copy the `Application (client) ID` and `Directory (tenant) ID` values,
   make them available as environment variables `OAUTH_CLIENT_ID`, `OAUTH_TENANT_ID`.
 7. Click on `Certificates & secrets` > `New client secret`.
 8. Fill in the form:
-   - Description: `Local-Intranet-v2`
+   - Description: `Staging-Intranet-v2`
```
wilson1000, line 342: Hey 😄 
A quick Q... why are we using v2, do we know what does it means? 
... maybe we can drop the v all together?

deploy/development/config.yml
```diff
@@ -9,5 +9,5 @@ data:
   WP_SITEURL: 'https://dev.intranet.justice.gov.uk/wp'
   AWS_CLOUDFRONT_HOST: 'cdn.dev.intranet.justice.gov.uk'
   # See Azure Setup for more information on how to get these values.
-  OAUTH_CLIENT_ID: "1dac3cbf-91d2-4c0e-9c80-0bf3f8fabd75"
-  OAUTH_TENANT_ID: "c6874728-71e6-41fe-a9e1-2e8c36776ad8"
+  OAUTH_CLIENT_ID: "51266573-d4eb-41db-8ebe-a9548aa4f01e"
+  OAUTH_TENANT_ID: "0bb413d7-160d-4839-868a-f3d46537f6af"
```
wilson1000, line 13: These might (will) get flagged by the security squad at some point. Do they need to be hidden or are we good to roll with them?

wilson1000 (commented):
Just some queries 👍

deploy/development/config.yml
EarthlingDavey: Ass discussed, I'll add a concise note as to why these are ok here.

.github/README.md
EarthlingDavey: Good point, it was to differentiate between Tactical Product's secrets, but I think it's unnecessary - I'll remove.

PR #724, opened 2024-10-08
Search: Remove page and documents from date based decay.

EarthlingDavey:


wilson1000 (approved):
This looks great @EarthlingDavey - let's run with it.

PR #731, opened 2024-10-09
CDPT 2144 Use a new crest of arms logo in the site header

wilson1000:
Review: https://staging.intranet.justice.gov.uk/

I have changed how the image is presented in the header due to the image size and dimensions. The image quality is poor, with too much detail and there may be an issue with auto-antialiasing in some views. I noticed the browser stops aliasing after some time, causing the image to look pixelated.


EarthlingDavey:
Hey @wilson1000 , I've seen your comment on the ticket.

What do you think about resizing the png to be 2x the width that it will be rendered at?

That way there should be no affect on the image due to browser resizing on retina screens.
And, on non-retina screens there should be at most 50% scaling.

I'm not certain, but I suspect the file size of a png will not be too bad compared to an svg.
If you think about the shape of the logo, with lots of curves and points, every change in radius and point will be a line of code in the svg.

I don't know if time/resources allow, but it would be interesting to see a size comparison.

EarthlingDavey (approved):

PR #739, opened 2024-10-15
Automate local secret key generation

wilson1000:


EarthlingDavey:
Looks really nice and works well here! :)

PR #741, opened 2024-10-16
Fix a bug in CloudFront signing and in fmp-status script.

EarthlingDavey:
For the status script, remove html from the query string, so that the output is in table layout.

For CloudFront, correctly decode the Policy cookie, from base64.

Prior to this fix, remainingTimeFromCookie was always returning 0, and issuing CF cookies on every request 🤦‍♂️


wilson1000 (approved):
Good spot @EarthlingDavey 😄

PR #742, opened 2024-10-16
CDPT-2186 Create a new container to push static assets as a k8s job.

EarthlingDavey:
This PR allows for theme, plugin and WordPress core assets to be synced to S3 and served via the CDN.

It can be broken down into parts/stages:

1. A build step to create a docker image that contains all of the assets.
2. A job is run at deploy time to run the docker image on Cloud Platform, this job syncs the assets to S3.
3. The application checks that the files are available via S3, if they ar the asset urls are re-written from https://intranet... to https://cdn.intranet...


EmilyHazlehurst (approved):
Looks good 😄

EarthlingDavey:
Awesome, ty 🤞

PR #744, opened 2024-10-17
CDPT-2079 Focus first new result after (infinite) pagination.

EarthlingDavey:


wilson1000 (approved):
Nice :) do we need to worry about the .append() function reinterpreting text as HTML?

EarthlingDavey:
Hey @wilson1000 , I've double and triple checked and I think it's ok. 

With `.append` and `.html` that have been flagged, the html content is trusted and has come from the server.

AFAICT, there is no opportunity for an attacker to get their own content into the parameters.

From my understanding this is bad:

```
# Get the id form the url, e.g. example.com#15
const pageNumber = window.location.hash;
# Make a html string
const html = `<p>You are on page number ${pageNumber}`
# Append the html
$("#content").append(html);
```

Because someone could visit example.com#<script>alert('pwned')</script> , and it would run the js.

But I think it's safe to do the following:

```
$.ajax({
  success: (response) => {
    // Transform the response... without using untrusted user inputs.
    // ...
    // Use .append with trusted html
    $("#content").append(response.html);
   }
});
```

wilson1000:
> But I think it's safe to do the following:

Hey @EarthlingDavey - can a user intercept the request before the server responds, or do we have complete trust that the server gave us the response and not another service?

I hope that made sense, and sorry for being (probably) catastrophic in my thinking!

EarthlingDavey:
I think it's possible that a user could manipulate requests, but all $_POST data is being sanitized in `content-filter/search.php` with `sanitize_text_field`.

This sanitization prevents 'reflected' xss attacks.

But, I've learned recently that 'stored' xss attacks can happen when a script tag that's stored in the database makes it's way into the DOM, because of insufficient escaping at render time.

I'll look into it, to see if this script is vulnerable to that kind of attack.

PR #745, opened 2024-10-17
Fixes the search filter panel on scroll when users are browsing.

wilson1000:


EarthlingDavey:
Hey @wilson1000 , I just spotted this PR even though you haven't requested a review.

I've done a similar thing on a different website, it might be worth checking to see if the css property `position: sticky` would be suitable here. If it's suitable it would mean you don't really need the custom js.

With a quick test in browser developer tools, adding the below css seems like it might work.

```
@media only screen and (min-width: 740px) {
    .l-main .l-secondary {
        ...
        position: sticky;
        top: 20px;
    }
}
```

PR #746, opened 2024-10-18
CDPT-2013 Show co-author-plus errors as admin notices - instead of 500.html

EarthlingDavey:
This PR shows the user any errors that the plugin co-author-plus generates.

This will assist in understanding the problem. Previously the user would have seen the generic 500.html error page.

image


wilson1000:
> This PR shows the user any errors that the plugin co-author-plus generates.
> 
> This will assist in understanding the problem. Previously the user would have seen the generic 500.html error page.
> 
> image

Hi @EarthlingDavey. Is the message text coming from Co-Authors Plus? 
I'd like us to remove the word WordPress from the `$message` string - is that possible?

EarthlingDavey:
Sure thing! It does come from the plugin, but I can filter the text.

PR #748, opened 2024-10-25
CDPT-2216 Reset global $post and WP_Query after getting recent.

EarthlingDavey:


public/app/themes/clarity/src/components/c-article/view.php
```diff
@@ -44,9 +47,12 @@
     <aside class="l-secondary">
     <?php
         echo '<h2 class="o-title">Recent blog posts</h2>';
-        foreach ($query->posts as $key => $post) {
+        $main_post = $post;
```
EmilyHazlehurst, line 50: @EarthlingDavey is $post not undefined initially here? I guess it's the Wordpress global $post but the reassignments here are a little odd - is there a cleaner way to do this?

Maybe something like:

```
$recent_query = new WP_Query($query_args->get());
...
if( $recent_query->have_posts()):
    while( $recent_query->have_posts()): $recent_query->the_post();
        include locate_template('src/components/c-article-item/view-blog-feed.php');
    endwhile;
endif;
wp_reset_postdata();
```

public/app/themes/clarity/src/components/c-news-article/view.php
```diff
@@ -37,9 +40,12 @@
...
     <?php
         echo '<h1 class="o-title">Recent news</h1>';
-        foreach ($query->posts as $key => $post) {
+        $main_post = $post;
+        foreach ($recent_query->posts as $key => $post) {
           include locate_template('src/components/c-article-item/view-news-feed.php');
         }
+        $post = $main_post;
+        unset($main_post);
     ?>
```
EmilyHazlehurst, line 49: Same comment as above.

public/app/themes/clarity/src/components/c-news-article/view.php
```diff
@@ -37,9 +40,12 @@
 
     <?php
         echo '<h1 class="o-title">Recent news</h1>';
-        foreach ($query->posts as $key => $post) {
+        $main_post = $post;
```
EmilyHazlehurst, line 43: Same question as for [c-article/view.php](https://github.com/ministryofjustice/intranet/pull/748/files#r1816478965)

public/app/themes/clarity/src/components/c-article/view.php
EarthlingDavey: Thanks for picking this up. Yes, basically you're 100% right :) Your example is the usual WordPress way of doing things.

It's good because it doesn't affect the global $post variable. It's even better if you use `get_template_part` instead of `locate_template` because the variables inside the template file are scoped. 

- We can control what goes in, with the 3rd arg of `get_template_part`
- And whatever variable are assigned in the template don't leak out.

Unfortunately `locate_template` is used throughout the theme, and I'm 99% sure this it quite a poor way to do things. I would like to hear some reasoning for it's use but I expect the dev who chose to do it that way is long gone.

---

I'll spend some time on trying to improve this, it will mean changing 2 loops in additional files, but that's not too excessive :)

EarthlingDavey:
Hey @EmilyHazlehurst , thanks for pushing back on the janky workaround. Would you mind re-reviewing? Thanks 🙌

wilson1000 (approved):
Looks good 😄

PR #755, opened 2024-10-29
Turn on debug log for all envs.

EarthlingDavey:
This is a reaction to seeing a 500 error on staging - having searched the logs, there was no indication of the cause.

This PR adds logging to `/dev/stderr`, it still does *not* display errors.


wilson1000 (approved):
Good thinking

PR #756, opened 2024-10-29
CDPT-2223 Update logo for heartbeat and test version modals

EarthlingDavey:


wilson1000:
Do you have a representation of the layout change, if there is one?

wilson1000:
@EarthlingDavey comment above, cheers mate 👍

EarthlingDavey:
Thanks for reviewing, there is no change to layout. 

I found myself resizing the image to get the paragraph text to position correctly, underneath the image.

Clearing float on the paragraph makes it always sit under the image.

wilson1000 (approved):
Nice mate... there is also a crest on the login page. We can do that later 😄 

Screenshot 2024-10-30 at 12 06 50

wilson1000:
One more thing, can we get a base64 representation of this crest, instead of loading it through the URL? We can maybe save this in CSS and call a class with sizes to show it. 
**What do you think**?

EarthlingDavey:
Good spot! Are you ok with me creating a new svg for now and revisiting changing it to base64 post migration?

wilson1000 (approved):

PR #757, opened 2024-10-29
CDPT-2228 Copying main theme to develop theme & various bug fixes.

EarthlingDavey:


public/app/themes/clarity/functions.php
```diff
@@ -81,6 +81,7 @@
 require_once 'inc/content-filter/search-query.php';
 require_once 'inc/content-filter/search.php';
 require_once 'inc/enqueue.php';
+require_once 'inc/environment-notice.php';
```
EarthlingDavey, line 84: I'm not sure if this was intentionally removed, I've assumed it was a mistake and added it back.

wilson1000 (commented):
Mate, just one comment about the environment notice.

public/app/themes/clarity/functions.php
EarthlingDavey: Ah ... thanks @wilson1000 !

PR #763, opened 2024-10-30
Make nginx readiness endpoint dependant on a working socket to fpm

EarthlingDavey:


public/metrics/service.php
```diff
@@ -140,6 +140,7 @@ public function serveMetrics(): void
     {
         header('Content-Type', 'text/plain');
         echo $this->getServiceMetrics();
+        unset($this->guzzle_client);
```
wilson1000, line 143: We tested this, but it had no visual effect. 
Do we still think it's needed, due to *good-housekeeping*, or should we consider removing?

wilson1000 (approved):
Hey @EarthlingDavey just a couple of comments. I'll leave the call to you.

public/metrics/service.php
EarthlingDavey: I don't think it can do any harm.

deploy/config/local/nginx/server.conf
```diff
@@ -303,8 +308,18 @@ server {
     # LOCATIONS - heath
     ##
 
-    location ~ ^/(liveness|readiness)$ {
+    location = /liveness {
```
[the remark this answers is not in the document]
EarthlingDavey: Good shout, I'll make the change.

PR #770, opened 2024-11-04
Bug/CDPT-2258 Handle empty post and text domain. Add constructor to notify-for-wordpress

EarthlingDavey:
There is a bug where the [notify dashboard](http://intranet.docker/wp/wp-admin/index.php?page=notify-dashboard) was firing a fatal php error.

This was because the plugin was not setting the text-domain correctly. This has been fixed in the upstream git repo and a patch fix in this PR.

In a defensive programming move, `coauthors_filter_text` has been changed so that text domain is an optional string.

`$post?->post_type` should also fix the warning `Attempt to read property "post_type" on null`


wilson1000 (approved):
Let's run this through to prevent the error. Do we have to keep an eye on this, to make sure an update doesn't break anything?

EarthlingDavey:
Good shout, I've made a note in the upstream PR that this find and replace must be removed if the upstream PR is merged and released.

PR #772, opened 2024-11-05
Create a new deployment that handles cron tasks

EarthlingDavey:
The aim of this PR is to see if we can determint the root cause of CPU spikes that are happening on production.

By isolating cron tasks to a dedicated deployment, we can more easily see if it is a scheduled task that is causing the CPU spikes.

Steps to create the new deployment:

- Duplicated the existing deployment.
- The cron container is removed from the original deployment.
- The new deployment is named and labeled as `${KUBE_NAMESPACE}-cron`


wilson1000 (approved):
👍

PR #775, opened 2024-11-05
Improve logging

EarthlingDavey:
- Log Relay (Redis) errors to stderr.
  A note on memory limit: https://relay.so/docs/1.x/configuration#memory-limits
- fpm to not log every request, it's makes the logs too noisy.
- Filter out known hosts from pre_http_request logging - to remove noise from the logs.
- Prevent API requests for core, theme and plugin versions.

### DIFF
--- a/public/app/themes/clarity/inc/updates.php
+++ b/public/app/themes/clarity/inc/updates.php
@@ -1,0 +1,67 @@
+<?php
+
+namespace MOJ\Justice;
+
+/**
+ * Functions to related to updates, and disabling checks.
+ */
+class Updates
+{
+    public function __construct()
+    {
+        $this->hooks();
+    }
+
+    /**
+     * @return void
+     */
+    public function hooks(): void
+    {
+        /**
+         * Prevent http requests to api.wordpress.org to check for WP core versions.
+         * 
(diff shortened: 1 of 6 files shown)


public/app/themes/clarity/inc/updates.php
```diff
@@ -0,0 +1,93 @@
...
+        /**
+         * Prevent http requests to api.wordpress.org to check for theme version.
+         * 
+         * This filter's function returns a dummy payload that will prevent an api request,
+         * and not cause an error.
+         * 
+         * @see https://wp-kama.com/2020/disable-wp-updates-check
+         * @see https://developer.wordpress.org/reference/functions/wp_update_themes/
+         */
+
```
wilson1000, line 51: Hey buddy, I was meant to mention this a while ago. My IDE prefers that the comment sits directly above the method, without a space in between. Are you happy (not now) in future to associate the comment directly with the function/method?

wilson1000 (approved):

public/app/themes/clarity/inc/updates.php
EarthlingDavey: Hey, absolutely, that's all good with me.

PR #777, opened 2024-11-07
CDPT-2264 Post go live improvements

EarthlingDavey:
This PR

- Fixes a [warning](https://ministryofjustice.sentry.io/issues/5686391577/?project=4506870789505024&query=is%3Aunresolved%20issue.priority%3A%5Bhigh%2C%20medium%5D&referrer=issue-stream&statsPeriod=14d&stream_index=4&utc=true) in cms-tree-page-view `Deprecated: htmlspecialchars_decode(): Passing null to parameter #1 `
- Replaces serialize/unserialize with igbinary_serialize/igbinary_unserialize It's faster and should have higher capacity.
- Adds phpredis and a config variable `WP_REDIS_USE_RELAY` to switch between it and Relay. Relay is [throwing errors](https://ministryofjustice.sentry.io/issues/?project=4506870789505024&query=is%3Aunresolved%20relay&referrer=issue-list&statsPeriod=14d&utc=true) and I suspect it could be the cause of CPU spikes.


.env.example
```diff
@@ -41,6 +41,7 @@ OPENSEARCH_URL=http://opensearch:9200
 
 CACHE_HOST=redis
 CACHE_SCHEME=redis
+# WP_REDIS_USE_RELAY=true
```
wilson1000, line 44: Don't we need this to be true?

deploy/production/config.yml
```diff
@@ -12,3 +12,4 @@ data:
   # The following IDs are not private, they form part of the publicly visible oauth login url.
   OAUTH_CLIENT_ID: "ffb808d2-312b-4ffe-a6e5-d6eacfd9f06f"
   OAUTH_TENANT_ID: "c6874728-71e6-41fe-a9e1-2e8c36776ad8"
+  WP_REDIS_USE_RELAY: 'true'
```
wilson1000, line 15: String or boolean?

wilson1000 (approved):

deploy/production/config.yml
EarthlingDavey: It must be a string, otherwise I get this in my editor linting.

> Incorrect type. Expected "string"

.env.example
EarthlingDavey: I'm thinking that Relay may be causing CPU spikes. With that in mind, it's optional, with default being unset.

I'm not certain, if you would prefer , I can change it.

PR #785, opened 2024-11-19
CDPT-2264 Add a timeout to the fpm readiness script, plus...

EarthlingDavey:
This PR means that even when the fpm server is not responding via the socket - the script will still `exit 1` rather than hang.

It also includes:

- Changes to Relay config, aiming to prevent the OOM error.
- A move of the `ewww-image-optimizer` plugin from `mu-plugins` so that we can trial disabling it.

### DIFF
--- a/bin/fpm-readiness.sh
+++ b/bin/fpm-readiness.sh
@@ -28,7 +28,7 @@ if [ $(echo "$PING_RESPONSE" | tail -1) != 'pong' ]; then exit 1; fi;
 
 
 # 2️⃣ Request the fpm status via fast-cgi.
-FPM_STATUS=$(env -i SCRIPT_NAME=/status SCRIPT_FILENAME=/status REQUEST_METHOD=GET cgi-fcgi -bind -connect /sock/fpm.sock);
+FPM_STATUS=$(env -i SCRIPT_NAME=/status SCRIPT_FILENAME=/status REQUEST_METHOD=GET cgi-fcgi -bind -connect /sock/fpm.sock &);
 
 # Store the process ID of the status request.
 STATUS_PID=$!;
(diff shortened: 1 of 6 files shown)


bin/fpm-readiness.sh
```diff
@@ -1,13 +1,53 @@
 #!/usr/bin/env ash
 
-# Send a ping request via fast-cgi. Use an empty environment with `env -i`.
-PING_RESPONSE=$(env -i SCRIPT_NAME=/ping SCRIPT_FILENAME=/ping REQUEST_METHOD=GET cgi-fcgi -bind -connect /sock/fpm.sock | tail -1);
+# 1️⃣ Send a ping request via fast-cgi. Use an empty environment with `env -i`.
+PING_RESPONSE=$(env -i SCRIPT_NAME=/ping SCRIPT_FILENAME=/ping REQUEST_METHOD=GET cgi-fcgi -bind -connect /sock/fpm.sock &);
```
wilson1000, line 4: Is this the Async issue I was referring to earlier? 🤔

bin/fpm-readiness.sh
```diff
@@ -1,13 +1,53 @@
...
+    # If the ping request is still running, terminate it.
+    kill $PING_PID;
+    # Print an error message.
+    echo "Ping request timed out.";
+    # Exit with code 1.
+    exit 1;
+fi;
+
+# Wait for the ping request process to complete.
+wait $PING_PID;
```
wilson1000, line 23: Nice that we can check the result 🚀 !

wilson1000 (approved):
Looks good 😄

wilson1000 (approved):
Nice!!

bin/fpm-readiness.sh
EarthlingDavey: I'm not sure... It's been a long day 🙃 I've added some notes to the Miro board.

We did talk about the potential for this script to be bringing down the container.

I tried to replicate a failure on staging and ran this script 10k times in 2 mins and it did not cause a CPU spike of the shape that we are trying to debug.

PR #793, opened 2024-11-21
CDPT-2264 Revert isolated cron deployments PR #773 and #772

EarthlingDavey:
Also:

- Partly revert #769 by decreasing min-replicas
- Remove CDPT_Debug logging from functions.php 
  (added in #785)


deploy/production/hpa.yml
```diff
@@ -8,7 +8,7 @@ spec:
     apiVersion: apps/v1
     kind: Deployment
     name: intranet-production
-  minReplicas: 6
+  minReplicas: 5
```
wilson1000, line 11: This would ideally be a number displaying the maximum number of replicas, divided by three:
`18 / 3 = 6`

I can't find a resource that supports this. However, when the engine runs and auto-scales, having a multiple of three is best.

deploy/production/deployment.tpl.yml
```diff
@@ -62,167 +62,6 @@ spec:
                   name: ${KUBE_NAMESPACE}-base64-secrets
                   key: IPS_FORMATTED
 
-        - name: fpm
```
wilson1000, line 65: FPM has been removed. Maybe you didn't mean to happen?

deploy/development/deployment.tpl.yml
```diff
@@ -62,167 +62,6 @@ spec:
                   name: ${KUBE_NAMESPACE}-base64-secrets
                   key: IPS_FORMATTED
 
-        - name: fpm
```
wilson1000, line 65: FPM has been removed. Maybe you didn't mean to happen?

wilson1000 (changes requested):
Some potential changes 👍

deploy/production/deployment.tpl.yml
EarthlingDavey: Hot dang! This would not be good! I think it's just the way GitHub is showing the diffs.

It looks ok in the [file view](https://github.com/ministryofjustice/intranet/blob/eb47431870d22d016c3df9d741c4b198cc2d20dd/deploy/production/deployment.tpl.yml).

deploy/production/hpa.yml
EarthlingDavey: I have a feeling that this was maybe the for K8S case a while ago, or there's chance you are thinking of the [split-brain problem](https://www.google.com/search?q=why+odd+number+of+nodes%3F&rlz=1C5GCEM_enGB1093GB1093&oq=why+odd+number+of+nodes%3F+&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIICAEQABgWGB4yCAgCEAAYFhgeMggIAxAAGBYYHjIKCAQQABiABBiiBDIKCAUQABiABBiiBDIKCAYQABiABBiiBDIKCAcQABiABBiiBDIKCAgQABiiBBiJBdIBCDY2MjVqMGo3qAIIsAIB&sourceid=chrome&ie=UTF-8) where consensus algorithms elect a leader.

---

If you'll humour me and unlearn the multiples of 3 rule for a moment, I think I might be able to explain.

The minimum replicas is a kind of baseline right? For arguments sake lets say it's the number of replicas that are needed for keeping the service running for 95% of the time - during normal operating conditions.

Maximum replicas is an upper limit based on a number of factors:

- how 'spikey' is the traffic.
- how many replicas would be needed to handle high traffic events like pay info being published.
- how far do you want to go to keep the service up and running under a DDOS attack.
- cost and available resources, setting it to 100 might be great to mitigate downtime during DDOS but it would affect other services in the cluster.

Both max and min can be determined by:

- analysis of historical data, like BAU traffic, high traffic events. 
- experiments with stress tests.

By applying a factor of 3 we are saying either:

1. Let's go with our best estimate of minimum but disregard our reasoning for maximum replicas, or
2. Lets set our best estimate for maximum replicas, and disregard our reasoning around what the minimum should be.

Do you see what I mean? Wouldn't it be better to say... we've looked at the data and set what we think is optimal for both min and max independently

deploy/production/deployment.tpl.yml
wilson1000: Haha, there must be a glitch with GitHub. 
They normally show where the code was added back 😄

EarthlingDavey: Haha, it has to be the most misleading GH diff I've ever seen.

wilson1000 (approved):

PR #797, opened 2024-11-26
Allow access to the intranet-archive via JWT.

EarthlingDavey:
This PR:

- Adds a WP CLI script for manual creation of a long-lived JWT for the intranet-archive scraper.
- Handles the newly created JWT role of `intranet-archive`, by allowing it access only via Cloud Platform egress IPs.


wilson1000 (approved):
Looks good dude 👍

PR #809, opened 2025-01-20
CDPT-2430 Remove basic auth from dev ingress

EarthlingDavey:


wilson1000:
Hey @EarthlingDavey , a few of other locations:
1. https://github.com/ministryofjustice/intranet/blob/main/deploy/development/secret.tpl.yml#L21
2. https://github.com/ministryofjustice/intranet/blob/main/.github/workflows/deploy.yml#L56
3. https://github.com/ministryofjustice/intranet/blob/main/.github/workflows/deploy.yml#L71-L85
4. https://github.com/ministryofjustice/intranet/blob/main/deploy/development/deployment.tpl.yml#L151-L155

Do we need to remove these? Or, would you prefer we think about this later?

EarthlingDavey:
Hey mate, good spot.

I hadn't thought much about it, but we are still using Basic Auth for the demo namespace.

I think I didn't make an Entra App for Demo because it's not being used at the moment.

---

Is it ok with you to send this through and work out what to do with demo next sprint?

wilson1000 (approved):

PR #811, opened 2025-01-23
POC revision of Entra Auth.

EarthlingDavey:
This POC is a change to the Entra Auth flow to use packages provided by Microsoft. In an aim to address an issue with Mobile Edge browsers.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
             "koodimonni/composer-dropin-installer": true,
             "composer/installers": true,
             "php-http/discovery": true,
-            "roots/wordpress-core-installer": true
+            "roots/wordpress-core-installer": true,
+            "tbachert/spi": false
         }
     },
     "repositories": [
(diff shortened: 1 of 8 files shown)


composer.json
```diff
@@ -10,7 +10,8 @@
             "koodimonni/composer-dropin-installer": true,
             "composer/installers": true,
             "php-http/discovery": true,
-            "roots/wordpress-core-installer": true
+            "roots/wordpress-core-installer": true,
+            "tbachert/spi": false
```
wilson1000, line 14: Buddy, what is this?

wilson1000 (approved):

composer.json
EarthlingDavey: Given that the direct dependency `microsoft/kiota-abstractions` is Microsoft, I expect their sub-dependencies will be regularly scanned for vulns. and be subject to scrutiny.

PR #813, opened 2025-03-05
CDPT-2482: Page margins

EmilyHazlehurst:
- Apply padding-top to all pages
- Reset the margins on l-full-page elements that are the first child of a page for consistency

### DIFF
--- a/public/app/themes/clarity/src/components/c-share-post/style.styl
+++ b/public/app/themes/clarity/src/components/c-share-post/style.styl
@@ -1,10 +1,6 @@
 // include the toolkit (a suite of variables, classes and mixins)
 @require "../../globals/css/_toolkit"
 
-.l-full-page{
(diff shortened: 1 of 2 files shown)


EarthlingDavey:
Hey @EmilyHazlehurst 👋 The code looks all good, I just noticed a minor change in the spacing between live and local.

image

It might just need a slight change to the values.

EarthlingDavey (approved):

EmilyHazlehurst:
Thanks @EarthlingDavey, I've removed the padding from pages that have breadcrumbs and from the l-main class if it's the child another l-main element.

wilson1000:
Hey team 😄 @EmilyHazlehurst @EarthlingDavey 
Thanks for your work on this. 
Let's keep working together to achieve consistency across the site 🚀

PR #817, opened 2025-03-07
CDPT-2265 wp-document-revisions - correct the most recent revision author

EarthlingDavey:
Reluctantly, I had to use composer-post-install.sh script to fix this bug.

The script adds 2 filters to wp-document-revisions functions.

These filters are then used in the theme to correct the author name in 2 places on the document edit screen.

### DIFF
--- a/public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
+++ b/public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
@@ -192,8 +192,11 @@ class WPDocumentRevisions
         }
 
         // If we don't have a post ID, but we are in the loop, get the post ID.
-        if (!$post_id && get_the_ID()) {
-            $post_id = get_the_ID();
+        $post_id = $post_id ?: get_the_ID();
+
+        // If we still don't have a post ID, return 0.
+        if (!$post_id) {
+            return 0;
         }
 
         // If we haven't already set the wp_document_revisions object, do so now.
@@ -202,7 +205,7 @@ class WPDocumentRevisions
         }
 
         // Get the revisions for the current post - the first in the array is the document, technically not a revision.
-        $document_revisions = $this->wp_document_revisions->get_revisions(get_the_ID());
+        $document_revisions = $this->wp_document_revisions->get_revisions($post_id);
 
         // In an edge case we might not have any revisions. If so, return 0.
         if (empty($document_revisions)) {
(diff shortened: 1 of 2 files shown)


public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
```diff
@@ -172,4 +176,102 @@ public function retryFilesNotFound($file, $attachment_id)
...
+     * Filter the get_revisions result to correct the author.
+     *
+     * When this filter is called in the context of 'revision_metabox',
+     * it is used to correct the author on the first row of the revisions table.
+     *
+     * @param array $revisions The revisions.
+     * @param string $context The context.
+     * @return array The filtered revisions.
+     */
+    public function filterGetMostRecentRevision($revisions, $context)
```
EmilyHazlehurst, line 238: Missing return type.

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
```diff
@@ -172,4 +176,102 @@ public function retryFilesNotFound($file, $attachment_id)
...
+     * Filter the get_latest_revision result to correct the author.
+     * 
+     * When this filter is called in the context of 'document_metabox',
+     * it is used to correct the author in the string 'Checked in x ago by y'.
+     * 
+     * @param mixed $revision The revision.
+     * @param string $context The context.
+     * @return mixed The filtered revision.
+     */
+    public function filterGetLatestRevision(mixed $revision, string $context): mixed
```
EmilyHazlehurst, line 263: Instead of mixed, could the $revision param type be WP_Post here?

EmilyHazlehurst (approved):
A couple of little comments, but otherwise looks good and works well 😄

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
EarthlingDavey: Hi thanks for the heads up. I've revised the types on filterGetMostRecentRevision & filterGetLatestRevision to better reflect the types that the upstream functions are returning.

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
EarthlingDavey: Thanks, I've used WP_Post in the docblock.

To reflect the code in wp-document-revisions, and to play it safe... I've removed the native php types - as if they are wrong, php will throw a fatal error.

PR #819, opened 2025-03-10
CDPT-2295 Add intranet-archive URL

EarthlingDavey:
This PR adds the intranet archive URL, and should be merged after the first snapshot has completed.


MeenaModhvadia (approved):
LGTM

PR #821, opened 2025-03-11
CDPT-2265 Remove persistent cache for `document_revisions` group

EarthlingDavey:


public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
```diff
@@ -212,20 +216,20 @@ public function getMostRecentRevisionAuthor(string $format = 'id', int|null $pos
             return 0;
         }
 
-        // Get the author of the most recent revision.
-        $most_recent_revision_author = $document_revisions[1]->post_author ?? null;
+        // Get the most recent revision, if there are no revisions, use the original document.
+        $most_recent = $document_revisions[1] ?? $document_revisions[0];
```
wilson1000, line 220: Can we protect `$document_revisions[0]` in the most unlikely case that `$document_revisions` is not an array?

```
$most_recent = $document_revisions[1] ?? ($document_revisions[0] ?? null);
```

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
```diff
@@ -212,20 +216,20 @@ public function getMostRecentRevisionAuthor(string $format = 'id', int|null $pos
...
 
         // If we can't find the author, return the 0;
-        if (!$most_recent_revision_author || !is_numeric($most_recent_revision_author)) {
+        if (!$most_recent?->post_author || !is_numeric($most_recent->post_author)) {
```
wilson1000, line 223: Should we also check if post_author is present for `is_numeric()` check?

Could we add NULL safe here?
```
if (!$most_recent?->post_author || !is_numeric($most_recent?->post_author)) {
```

wilson1000 (changes requested):
A couple of comments for review 👍

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
wilson1000: It could be `false`

https://github.com/wp-document-revisions/wp-document-revisions/blob/39b774f843e7b957c129fb90f8acbdca9dfc44bc/includes/class-wp-document-revisions.php#L931

public/app/themes/clarity/inc/admin/plugins/wp-document-revisions.php
EarthlingDavey: Hey, thanks, I was thinking that in a case where `$most_recent->post_author` is not set, the first part of the if statement will be true, and the part after the `||` operator will never be run.

The part after the `||` operator will only be executed if `$most_recent->post_author` is truthy, so I don't think we need the `?->` symbol there.

wilson1000: Ah. Thanks for explaining!

wilson1000 (approved):
Looks good 👍

PR #823, opened 2025-03-12
CDPT-2570: HMI Probation / Youth Custody links

EmilyHazlehurst:
- Add clarification text to hmi link to reduce requests for access


EarthlingDavey (approved):
LGTM 🚀

PR #827, opened 2025-03-13
S3-Pusher ASCII Art

wilson1000:
Defines the S3 sync container using ASCII Art in the docker file


EarthlingDavey (approved):
Looks good! 💯

PR #829, opened 2025-03-24
CDPT-2601 Downgrade to WP Media Offload Lite

EarthlingDavey:
This PR:

- Removes WP Media Offload Pro from the mu-plugins folder, to plugins.
- Installs WP Media Offload Lite to the plugins folder.
- Adds code to the fpm container startup script, to ensure at least one of the WP Media Offload plugins is active.
- Adds theme compatibility for the Lite version.


wilson1000 (approved):
Looks good

PR #835, opened 2025-04-01
Remove composer.wp.dsd.io dependency

EarthlingDavey:
This PR:
- removes the dependency on composer.wp.dsd.io
- pulls 2 MOJ repos directly from GitHub


EarthlingDavey:
Thanks @wilson1000 - feels good to do some housekeeping!

PR #837, opened 2025-04-02
Update wp-admin-bar.php

wilson1000:
Extrapolates role formatting, as the role is only used once but might be useful in other areas.


public/app/themes/clarity/inc/admin/wp-admin-bar.php
```diff
@@ -96,6 +88,23 @@ function agency_context_switcher_menu($wp_admin_bar)
     );
 }
 
+function get_agency_user_role()
```
EarthlingDavey, line 91: This is just a nit, but as there is nothing specific related to agency in the function, would `get_formatted_user_role` be a better fit?

EarthlingDavey (approved):
Looks good 👍 I've left a couple of comments...

PR #840, opened 2025-04-09
CDPT-2623 Create Notes from Amy

EarthlingDavey:


wilson1000 (approved):
Looks good @EarthlingDavey let's see it in the wild 😄

PR #841, opened 2025-04-11
CDPT-2608 Add audit logs

EarthlingDavey:
This PR adds an auditing plugin. The basic functionality is there, but a follow up ticket should be created to add missing functionality of tracking ACF updates.


wilson1000 (approved):
I love what you've done here @EarthlingDavey. The security forethought is spot on 🚀

PR #845, opened 2025-05-20
Housekeeping

EarthlingDavey:


public/app/themes/clarity/src/vendors/magnific-popup.js
```diff
@@ -331,7 +331,10 @@
       $('html').css(windowStyles);
 
       // add everything to DOM
-      mfp.bgOverlay.add(mfp.wrap).prependTo(mfp.st.prependTo || $(document.body));
+      // Passing `mfp.st.prependTo` as an agrgument is flagging a security vulnerability
+      // Since we are not using that feature, we can safely remove it
+      // mfp.bgOverlay.add(mfp.wrap).prependTo(mfp.st.prependTo || $(document.body));
+      mfp.bgOverlay.add(mfp.wrap).prependTo($(document.body));
```
wilson1000, line 337: Hey, I don't think we use this library in the Intranet. Can we check?

EarthlingDavey: It's used in the codebase for the component `c-gallery-lightbox`, and that's used int the template `page_media_grid.php`. I've checked the database, it's still used on a handful of pages.

wilson1000 (approved):

PR #849, opened 2025-06-04
Scale unused demo environment to zero `demo/deployment.tpl.yml`

earthlingdavey:
As demo isn't regularly deployed, it is running old docker images, and these will contain outdated dependencies.

Since we are not currently using the demo environment, let's scale it to zero.


wilson1000:
Hey mate, who decided on this?

EarthlingDavey:
Hey Damien, I'm just taking some initiative with the aim to secure our deployments.

We have take lots of precautions based on security alerts, but we also need to make sure we aren't running old code, right?

PR #851, opened 2025-06-04
Redirect traffic for jac.intranet.service.justice.gov.uk (part 2)

EarthlingDavey:


wilson1000 (commented):
Hi @EarthlingDavey I'm not sure this is necessary. We are defining the domain as a static resource, which is not Important any longer. It has been discontinued. Is this another case of initiative or has the decision to keep the domain been overruled somewhere?

To add, the redirect is working as intended.

EarthlingDavey:
@wilson1000 , I applied these changes manually yesterday - since the redirect doesn't work without them, and I didn't want to degrade any service.

It is part of the plan as in this ticket: https://github.com/ministryofjustice/cloud-platform-environments/pull/33736#issue-3117419283

wilson1000:
Thanks for the explanation, it makes sense; however, the domain was getting decommissioned. The effort here has, in effect, brought it back into use. 

How long do you think we will need this in place? I would have suggested we remove it completely since the service has been migrated for some time now.

PR #852, opened 2025-06-09
CDPT-2653 Prevent edit and deletion of key pages, by non-admins.

earthlingdavey:


public/app/themes/clarity/inc/admin/users/protect-key-pages.php
```diff
@@ -0,0 +1,119 @@
...
+     *     @type int       $1 Concerned user ID.
+     *     @type mixed  ...$2 Optional second and further parameters, typically object ID.
+     * }
+     * @param WP_User  $user    The user object.
+     * @return bool[]           Filtered array of capabilities.
+     */
+    public function filterUserCapabilities($allcaps, $cap, $args)
+    {
+        // Check if the user is an administrator
+        if (isset($allcaps['administrator']) && $allcaps['administrator']) {
```
EarthlingDavey, line 90: This can be tested locally by commenting out this block.

Or you can, switch users to a non-admin user.

public/app/themes/clarity/inc/admin/users/protect-key-pages.php
```diff
@@ -0,0 +1,119 @@
...
+            return $allcaps; // Admins have all capabilities
+        }
+
+        // The second argument is typically the object ID (post ID, page ID, etc.)
+        if (!isset($args[2]) || !is_numeric($args[2])) {
+            // The second argument is not set or is numeric, so do nothing.
+            return $allcaps;
+        }
+
+        if (!$this->isKeyPage((int) $args[2])) {
```
wilson1000, line 100: Error checking is being performed on line 95 for `$args[2]`, but error checking has been omitted here. We should still check that array position 2 is still set.

public/app/themes/clarity/inc/admin/users/protect-key-pages.php
```diff
@@ -0,0 +1,119 @@
...
+            // If we are not dealing with a key page, do nothing.
+            return $allcaps;
+        }
+
+        // If we are dealing with a key page, and WP is checking if the user can edit it.
+        if ('edit_post' === $args[0]) {
```
wilson1000, line 106: Also relevant for line 111. 
Even though `$args[0]` is a required parameter, it would be safer to check that the key exists.

wilson1000 (commented):
Hey @EarthlingDavey, the PR looks great! Well done. Just a couple of comments.

public/app/themes/clarity/inc/admin/users/protect-key-pages.php
EarthlingDavey: Hey mate, thanks for the review :)

Line 97 has an early return on it, so if $args[2] is not set, there will be an early return and we won't get to this line.

So I think another check here will be unnecessary.

PR #857, opened 2025-06-16
Create integration-qa.yml

wilson1000:
This action is active for PRs and fires when the PR is ready for review. If the PR is in draft state, CI won't take place.

### DIFF
--- a/.github/workflows/integration-qa.yml
+++ b/.github/workflows/integration-qa.yml
@@ -1,0 +1,61 @@
+name: Deploy to QA Environment
+
+on:
+  pull_request:
+    types:
+      - converted_to_draft
+      - opened
+      - ready_for_review
+      - reopened
+      - synchronize
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  eligibility:
+    name: Determine eligibility
+    if: github.event.ref != 'refs/heads/main'
+    runs-on: ubuntu-latest
+    outputs:
+      eligible: ${{ steps.eligible.outputs.result }}
+    steps:
+      - name: Get draft PRs
+        id: get-draft-prs
+        env:
+          GH_TOKEN: ${{ secrets.ACTIONS_TOKEN }}
+        run: |
+          echo "result=$(gh pr list --repo ministryofjustice/intranet --draft --state open --json number)" >> "$GITHUB_OUTPUT";
+      - name: Verify PR status
+        id: eligible
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const draftPRs = ${{ steps.get-draft-prs.outputs.result }}.map( ( pr ) => pr.number );
+            return ! draftPRs.includes( ${{ github.event.pull_request.number }} );
+  image:
+    name: "Image"
+    needs: eligibility
+    if: ${{ needs.eligibility.outputs.eligible == 'true' }}
+    uses: ./.github/workflows/build.yml
+    secrets: inherit
+
+  modsec_config:
+    name: "Modsec"
+    needs: eligibility
+    if: ${{ needs.eligibility.outputs.eligible == 'true' }}
+    uses: ./.github/workflows/modsec-config.yml
+    secrets: inherit
+
+  deploy_qa:
+    name: "Quality Control"
+    if: ${{ needs.eligibility.outputs.eligible == 'true' }}
+    uses: ./.github/workflows/deploy.yml
+    needs: [eligibility, image, modsec_config]
+    with:
+      environment: demo
+      registry: ${{ needs.image.outputs.registry }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -16,10 +16,21 @@ jobs:
...
       id-token: write
       contents: read
+      pull-requests: write
     steps:
       - name: "Checkout"
         uses: actions/checkout@v4
 
+      - name: "Comment PR to indicate build start"
+        if: github.event_name == 'pull_request'
+        uses: thollander/actions-comment-pull-request@v3
```
earthlingdavey, line 26: Are we ok with using this? And, for it to have the `id-token: write` permission?

I started to review the code for this action, but [one js file](https://github.com/thollander/actions-comment-pull-request/blob/main/lib/index.js) is > 30k LOC

earthlingdavey:
@wilson1000 I merged in a couple of changes to disable Entra Auth for demo.

Would you mind reviewing them, some of it is borrowed from Justice UK and I've added something to handle the heartbeat endpoint.

Your work looks very good! I just had a few questions about security.

.github/workflows/build.yml
earthlingdavey: Yeah if it could be more isolated to a job or file, with just `pull-requests: write` then it would be minimise the attack surface.

.github/workflows/integration-qa.yml
```diff
@@ -0,0 +1,71 @@
...
+    runs-on: ubuntu-latest
+    outputs:
+      eligible: ${{ steps.eligible.outputs.result }}
+    steps:
+      - name: Get draft PRs
+        id: get-draft-prs
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          echo "result=$(gh pr list --repo ministryofjustice/intranet --draft --state open --json number)" >> "$GITHUB_OUTPUT";
```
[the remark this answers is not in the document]
earthlingdavey: Awesome!

earthlingdavey (approved):

.github/workflows/build.yml
wilson1000: Do you know what we are using `id-token` for?

.github/workflows/integration-qa.yml
```diff
@@ -0,0 +1,73 @@
...
+    secrets: inherit
+
+  modsec_config:
+    name: "Modsec"
+    needs: eligibility
+    if: ${{ needs.eligibility.outputs.eligible == 'true' }}
+    uses: ./.github/workflows/modsec-config.yml
+    secrets: inherit
+
+  deploy_qa:
```
earthlingdavey, line 63: Due to the ECR policy, 5 runs of this could knock the production image off of the ECR registry.

https://github.com/ministryofjustice/cloud-platform-environments/blob/main/namespaces/live.cloud-platform.service.justice.gov.uk/intranet-production/resources/ecr.tf

.github/workflows/deploy.yml
```diff
@@ -161,3 +163,12 @@ jobs:
       - name: "Apply manifest files"
         run: |
           kubectl -n ${KUBE_NAMESPACE} apply -f deploy/${{ inputs.environment }}
+
```
earthlingdavey, line 166: Do we need an extra `kubectl ` command here to wait for the deployment to be ready?

.github/workflows/deploy.yml
```diff
@@ -161,3 +163,12 @@ jobs:
...
+      - name: "QA Link"
+        if: inputs.environment == 'demo'
+        uses: thollander/actions-comment-pull-request@v3
```
earthlingdavey, line 169: The security improvement looks good!

If `thollander/actions-comment-pull-request` is compromised, it could replace links to a phishing website.

Is that an acceptable risk?

wilson1000: Thanks for clarifying mate! Well, we could perform a v-scan on the image before pushing it. Would that help us?

.github/workflows/integration-qa.yml
wilson1000: Thanks. Super-happy you found this, buddy! 
Let's chat about ways to solve this one.

.github/workflows/deploy.yml
wilson1000: I don't think so 🤔  what do you think?

.github/workflows/deploy.yml
earthlingdavey: A scan might help. An alternative could be to review the code and pin the version by hash.

There is an article that explains the problem and why pinning is one solution 

https://blog.rafaelgss.dev/why-you-should-pin-actions-by-commit-hash

> Pinned dependencies help reduce various security risks:
> ...
> - They can help mitigate compromised dependencies from compromising the project’s security. By evaluating the pinned dependency and being confident that it is not compromised, you can prevent the use of a later version that may be compromised.

PR #862, opened 2025-06-23
CDPT-1709 Promote deployment config from dev to staging & prod.

earthlingdavey:
@wilson1000 , this is working on dev, but only with one pod.

There is an issue where other pods cannot be looked up by their name.

I assumed that pods would have DNS records, like services, but they don't.

I'm changing the code so that we save pod IP addresses instead of pod hostnames.

### DIFF
--- a/.github/workflows/php-composer-test.yml
+++ b/.github/workflows/php-composer-test.yml
@@ -1,4 +1,4 @@
-name: Composer Test
+name: Composer
 
 on: [pull_request]
 
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        php-version: ['8.2', '8.3']
+        php-version: ['8.3', '8.4']
 
     steps:
     - uses: actions/checkout@v4


deploy/production/deployment.tpl.yml
```diff
@@ -163,6 +169,16 @@ spec:
                 secretKeyRef:
                   name: elasticache-output
                   key: auth_token
+            # This is the name of the pod, set by Kubernetes.
+            # Pods on the network can use this hostname to communicate with each other.
+            # i.e. An nginx purge url can be created with <nginx_hostname>:8080/purge/<url>
+            #      Purge urls for all pods, can be called from a single fpm container.
+            # N.B. On Kubernetes, the hostname is the same for all containers in the pod.
+            #      On Docker Compose, the hostname is the name of the service.
+            - name: NGINX_HOSTNAME
```
wilson1000, line 178: If this is a variable, will we need to wrap it with ${} ?

wilson1000 (approved):

deploy/production/deployment.tpl.yml
earthlingdavey: Good point, it is a variable name, but we don't want it to be expanded during the GH deploy action - so it's ok to keep as-is.

PR #865, opened 2025-06-25
Fix #843 - remove amazon-s3-and-cloudfront-pro logging

### LINKED ISSUE #843: Plugin file doesn't exist
https://github.com/ministryofjustice/intranet/blob/e39a955e4a1f06a1ad1257f557251c05dbfc95ec/bin/composer-post-install.sh#L18

earthlingdavey:
Logging, that was important during migration to Cloud Platform, has been removed from the composer post install script.


wilson1000 (approved):
👍

PR #888, opened 2025-09-02
CDPT-3003 Return `WP_Error` instead of throwing it. Handle documents without attachments.

earthlingdavey:
This PR fixes an error and a warning that became apparent when running the code on the dev. environment. 

- `return new WP_Error(...)` is used instead of `Throw new WP_Error(...)`.
- `is_wp_error($data)` has been used in favour of `try catch`.
- Documents without attachments are skipped in the loop, with `if (!$wpdr->get_document($document->ID))`.

### DIFF
--- a/public/app/themes/clarity/inc/api/synergy-feed-api/synergy-feed-api.php
+++ b/public/app/themes/clarity/inc/api/synergy-feed-api/synergy-feed-api.php
@@ -136,10 +136,9 @@ class SynergyFeedApi
      * It returns an array containing the matching pages.
      * 
      * @param WP_REST_Request $request The request object containing the parameters.
-     * @return array The response data containing the feed items.
-     * @throws WP_Error If the page is not found or if there is an error in the request.
+     * @return array|WP_Error The response data containing the feed items.
      */
-    public function getFeed(WP_REST_Request $request): array
+    public function getFeed(WP_REST_Request $request): array|WP_Error
     {
         $format = $request->get_param('format');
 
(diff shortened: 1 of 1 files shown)


public/app/themes/clarity/inc/api/synergy-feed-api/synergy-feed-api.php
```diff
@@ -231,11 +230,19 @@ public function getFeed(WP_REST_Request $request): array
...
                 if (!$document->post_status || $document->post_status !== 'publish') {
-                    continue; // If the document is not published, skip to the next iteration.
+                    // If the document is not published, skip to the next iteration.
+                    continue;
+                }
+
+                if (!$wpdr->get_document($document->ID)) {
+                    // If the document is not found in WP Document Revisions, skip to the next iteration.
+                    // This is an edge case where a document has no attachment.
+                    continue;
```
wilson1000, line 245: Looks good, buddy. 
Here's an alternative approach using a concise measure, with less repetition.

```php
// Skip to the next iteration, if the document...
switch(true) {
      // ... does not exist
    case (!$document): 
      // ... is not published
    case (!$document->post_status || $document->post_status !== 'publish'):
      // ... is not found in WP Document Revisions (edge case where a document has no attachment)
    case (!$wpdr->get_document($document->ID)):
        continue;
}
```

wilson1000 (approved):

public/app/themes/clarity/inc/api/synergy-feed-api/synergy-feed-api.php
earthlingdavey: Agreed, it is a bit wordy 😅 Will take your suggestion on board, thanks :)

PR #898, opened 2025-10-23
CDPTKAN-287 Update packages (amend to require env file)

earthlingdavey:
This fixes an oversight in the previous PR, where the env function was not available to the metrics class.


wilson1000 (approved):
🚀

PR #1066, opened 2026-07-02
Update code around document exports

earthlingdavey:


brown-a2 (approved):
What do we use this command line for? Is it integrated into something else or just a command line for use when we need to find something?

earthlingdavey:
@brown-a2 just a standalone / manual script.

PR #1067, opened 2026-07-02
CDPTKAN-287 Auth, remove callback version 2 that was previously used for debugging.

earthlingdavey:


earthlingdavey:
Already merged

PR #1068, opened 2026-07-02
Dev branch deploys to dev environment.

earthlingdavey:
This PR includes 2 changes:

- An update to the deployment workflow, so that the dev branch deploys to dev environment.
- A helper workflow to automate the syncing of dev branch with main. This new workflow should enable developers to bring dev in sync, in case it diverges from main. The workflow can be triggered by the Actions tab - and it will create a PR.


brown-a2 (approved):
LGTM. Maybe you could take us through this in a bit more detail one day? Also I like the sync dev with main. Maybe we add this to hale-platform?

PR #1081, opened 2026-07-10
Update WordPress to 7.0.1 and use official container image

earthlingdavey:
The idea with this PR is to use the official WordPress image, instead of our custom base image.

I have kept the /public/wp/ folder structure since it has benefits, and is high effort to change. e.g. Update nginx location blocks, the way composer installs packages. However, this could be done in the future as a separate ticket.

In this PR:

- Update WordPress to 7.0.1
- Use official container image (borrowing parts of hale-platform Dockerfile)
- Clean out IE styles
- Keep previous entrypoint file, and populate /wp from the wordpress image (instead of via composer)
- Disable AI support
- Improve error handling (for quieter logs)
- Update jQuery - away from deprecated functions (for quieter console logs)
- Move `homepage_all_access` capabilities functions out of hot-path to startup script
- Fix infinite loop when unrecognised agency cookie is present


public/app/themes/clarity/functions.php
```diff
@@ -39,7 +39,7 @@
 require_once 'inc/admin/permission-display-page.php';
 require_once 'inc/admin/plugins/cms-tree-page-view.php';
 require_once 'inc/admin/plugins/co-authors-plus.php';
-require_once 'inc/admin/plugins/polls.php';
+// require_once 'inc/admin/plugins/polls.php';
```
brown-a2: is it worth just removing completely?

brown-a2 (approved):

public/app/themes/clarity/functions.php
earthlingdavey: Good call

PR #1109, opened 2026-07-28
Update composer dependencies

earthlingdavey:


RobLoweMOJ (approved):
LGTM, though as discussed be good to update acf aswell.

PR #1143, opened 2026-09-02
Fix elasticpress cache poisoning - update to 5.3.4

earthlingdavey:
elasticpress 5.3.3 introduced a bug where a partial result (without patent ID) for a post would be cached in object cache.

This would surface as a bug where search result URLs were missing the `/parent/pages/` part.

This update fixes that bug.


RobLoweMOJ (approved):
LGTM

PR #1157, opened 2026-09-15
ACF upgrade - ahead of WP Core 7.1

earthlingdavey:


RobLoweMOJ (approved):
LGTM though i would implement copilots suggestion

